PR #1, opened 2020-07-21
Add initial config files

YevgeniyShunevych:


.editorconfig
```diff
@@ -0,0 +1,242 @@
+# Common .NET conventions, code formatting and naming convention rules. Check out possible configs here:
```
Piedone, line 4: This is exactly the same as in the dev branch of the Finitive solution, right?

.editorconfig
```diff
@@ -0,0 +1,242 @@
+# Common .NET conventions, code formatting and naming convention rules. Check out possible configs here: 
+# https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
+# If you found some issue then try to fix it, which you can also do from VS Quick Actions: 
+# https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-language-conventions?view=vs-2019#automatically-configure-code-styles
+# However, in this case PLEASE LET US KNOW in the Developers channel since such conventions should be exactly the same
+# for all solutions of ours.
```
Piedone: This can now be removed.

Readme.md
```diff
@@ -9,7 +9,27 @@
...
+- [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/)
+- [StyleCop.Analyzers](https://www.nuget.org/packages/StyleCop.Analyzers/)
+- [SonarAnalyzer.CSharp](https://www.nuget.org/packages/SonarAnalyzer.CSharp/)
+
+### How to add analyzers to project repository
+
+1. Add to `.gitmodules` file:
+   ```
+   [submodule "analyzers"]
+       path = analyzers
```
Piedone: Let's call it "Lombiq.Analyzers" so it's in sync with other projects of ours.

Readme.md
```diff
@@ -9,7 +9,27 @@
...
+       url = https://github.com/Lombiq/.NET-Analyzers.git
+       branch = dev
+   ```
+1. Create `Directory.Build.props` file in the repository root folder with the following content:
```
Piedone: This can be a folder up from the solution file as well, right?

General.ruleset
```diff
@@ -0,0 +1,110 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Piedone, line 1: I get such warnings:

image

```
Severity	Code	Description	File	Detail Description
Warning	AD1000	An instance of analyzer Microsoft.NetCore.CSharp.Analyzers.Runtime.CSharpDoNotRaiseReservedExceptionTypesAnalyzer cannot be created from D:\USERS\ZOLTÁN\APPDATA\LOCAL\MICROSOFT\VISUALSTUDIO\16.0_1C3E7029\EXTENSIONS\F1RDSMCY.WKK\Microsoft.NetCore.CSharp.Analyzers.dll: Could not load type 'Microsoft.NetCore.CSharp.Analyzers.Runtime.CSharpDoNotRaiseReservedExceptionTypesAnalyzer' from assembly 'Microsoft.NetCore.CSharp.Analyzers, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'..		System.TypeLoadException: Could not load type 'Microsoft.NetCore.CSharp.Analyzers.Runtime.CSharpDoNotRaiseReservedExceptionTypesAnalyzer' from assembly 'Microsoft.NetCore.CSharp.Analyzers, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.
   at System.Reflection.RuntimeAssembly.GetType(RuntimeAssembly assembly, String name, Boolean throwOnError, Boolean ignoreCase, ObjectHandleOnStack type)
   at System.Reflection.RuntimeAssembly.GetType(String name, Boolean throwOnError, Boolean ignoreCase)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerFileReference.Extensions`1.GetAnalyzersForTypeNames(Assembly analyzerAssembly, IEnumerable`1 analyzerTypeNames, Boolean& reportedError)
-----

Warning	AD1000	An instance of analyzer Microsoft.NetCore.Analyzers.Security.DoNotDisableSchUseStrongCrypto cannot be created from D:\USERS\ZOLTÁN\APPDATA\LOCAL\MICROSOFT\VISUALSTUDIO\16.0_1C3E7029\EXTENSIONS\F1RDSMCY.WKK\Microsoft.NetCore.Analyzers.dll: Could not load type 'Microsoft.NetCore.Analyzers.Security.DoNotDisableSchUseStrongCrypto' from assembly 'Microsoft.NetCore.Analyzers, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'..		System.TypeLoadException: Could not load type 'Microsoft.NetCore.Analyzers.Security.DoNotDisableSchUseStrongCrypto' from assembly 'Microsoft.NetCore.Analyzers, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.
   at System.Reflection.RuntimeAssembly.GetType(RuntimeAssembly assembly, String name, Boolean throwOnError, Boolean ignoreCase, ObjectHandleOnStack type)
   at System.Reflection.RuntimeAssembly.GetType(String name, Boolean throwOnError, Boolean ignoreCase)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerFileReference.Extensions`1.GetAnalyzersForTypeNames(Assembly analyzerAssembly, IEnumerable`1 analyzerTypeNames, Boolean& reportedError)
-----

Warning	AD1000	An instance of analyzer Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines.CSharpRemoveEmptyFinalizersAnalyzer cannot be created from D:\USERS\ZOLTÁN\APPDATA\LOCAL\MICROSOFT\VISUALSTUDIO\16.0_1C3E7029\EXTENSIONS\TUNWRTQO.YS2\Microsoft.CodeQuality.CSharp.Analyzers.dll: Could not load type 'Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines.CSharpRemoveEmptyFinalizersAnalyzer' from assembly 'Microsoft.CodeQuality.CSharp.Analyzers, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'..		System.TypeLoadException: Could not load type 'Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines.CSharpRemoveEmptyFinalizersAnalyzer' from assembly 'Microsoft.CodeQuality.CSharp.Analyzers, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.
   at System.Reflection.RuntimeAssembly.GetType(RuntimeAssembly assembly, String name, Boolean throwOnError, Boolean ignoreCase, ObjectHandleOnStack type)
   at System.Reflection.RuntimeAssembly.GetType(String name, Boolean throwOnError, Boolean ignoreCase)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerFileReference.Extensions`1.GetAnalyzersForTypeNames(Assembly analyzerAssembly, IEnumerable`1 analyzerTypeNames, Boolean& reportedError)
-----

```

It's most possibly related to this what I see: I don't get the solution-wide results reliably in the Error List. The analysis rules work all the time when I open a file, i.e. I see warnings corresponding to that file. However, for the whole solution, I just see them sporadically. I've seen them twice after a build before but not anymore and I can't really correlate it to anything: Clean-build doesn't help, reopening the solution neither, having a file open or not doesn't have an effect, and just waiting doesn't help either.

Also, this is likely behind me getting a VS exception when I hit Ctrl+. on a squiggly line:

```
System.TypeLoadException: Could not load type &apos;Microsoft.CodeQuality.CSharp.Analyzers.ApiDesignGuidelines.CSharpEnumsShouldHavePluralNamesFixer&apos; from assembly &apos;Microsoft.CodeQuality.CSharp.Analyzers, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35&apos;.&#x000D;&#x000A;   at System.Reflection.RuntimeModule.GetType(RuntimeModule module, String className, Boolean ignoreCase, Boolean throwOnError, ObjectHandleOnStack type)&#x000D;&#x000A;   at System.Reflection.RuntimeModule.GetType(String className, Boolean throwOnError, Boolean ignoreCase)&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.Reflection.TypeRef.get_ResolvedType()&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.Reflection.ResolverExtensions.Resolve(MethodRef methodRef)&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.Reflection.MethodRef.Resolve()&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.RuntimePartLifecycleTracker.CreateValue()&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.Create()&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveNext(PartLifecycleState nextState)&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveToState(PartLifecycleState requiredState)&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.GetValueReadyToExpose()&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.&lt;&gt;c__DisplayClass15_0.&lt;GetExportedValueHelper&gt;b__0()&#x000D;&#x000A;--- End of stack trace from previous location where exception was thrown ---&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.&lt;&gt;c__DisplayClass15_0.&lt;GetExportedValueHelper&gt;b__0()&#x000D;&#x000A;   at Microsoft.VisualStudio.Composition.DelegateServices.&lt;&gt;c__DisplayClass2_0`1.&lt;As&gt;b__0()&#x000D;&#x000A;   at System.Lazy`1.CreateValue()&#x000D;&#x000A;   at System.Lazy`1.LazyInitValue()&#x000D;&#x000A;   at System.Lazy`1.get_Value()&#x000D;&#x000A;   at Microsoft.CodeAnalysis.CodeFixes.CodeFixService.&lt;&gt;c__DisplayClass31_1.&lt;GetFixerPerLanguageMap&gt;b__0()&#x000D;&#x000A;   at System.Lazy`1.CreateValue()&#x000D;&#x000A;--- End of stack trace from previous location where exception was thrown ---&#x000D;&#x000A;   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()&#x000D;&#x000A;   at System.Lazy`1.get_Value()&#x000D;&#x000A;   at Microsoft.CodeAnalysis.CodeFixes.CodeFixService.&lt;AppendFixesAsync&gt;d__19.MoveNext()&#x000D;&#x000A;--- End of stack trace from previous location where exception was thrown ---&#x000D;&#x000A;   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)&#x000D;&#x000A;   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)&#x000D;&#x000A;   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)&#x000D;&#x000A;   at Microsoft.CodeAnalysis.CodeFixes.CodeFixService.&lt;GetFixesAsync&gt;d__16.MoveNext()&#x000D;&#x000A;--- End of stack trace from previous location where exception was thrown ---&#x000D;&#x000A;   at Roslyn.Utilities.TaskExtensions.WaitAndGetResult_CanCallOnBackground[T](Task`1 task, CancellationToken cancellationToken)&#x000D;&#x000A;   at Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionsSourceProvider.SuggestedActionsSource.GetCodeFixes(ITextBufferSupportsFeatureService supportsFeatureService, ISuggestedActionCategorySet requestedActionCategories, Workspace workspace, Document document, SnapshotSpan range, Func`2 addOperationScope, CancellationToken cancellationToken)&#x000D;&#x000A;   at Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionsSourceProvider.SuggestedActionsSource.GetSuggestedActions(ISuggestedActionCategorySet requestedActionCategories, SnapshotSpan range, IUIThreadOperationContext operationContext, CancellationToken cancellationToken)&#x000D;&#x000A;   at Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionsSourceProvider.SuggestedActionsSource.GetSuggestedActions(ISuggestedActionCategorySet requestedActionCategories, SnapshotSpan range, IUIThreadOperationContext operationContext)&#x000D;&#x000A;   at Microsoft.VisualStudio.Language.Intellisense.Implementation.LightBulbSession.&lt;&gt;c__DisplayClass40_4.&lt;TryGetSuggestedActionSets&gt;b__1()&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Utilities.GuardedOperations.CallExtensionPoint[T](Object errorSource, Func`1 call, T valueOnThrow, Predicate`1 exceptionToIgnore, Predicate`1 exceptionToHandle)&#x000D;&#x000A;--- End of stack trace from previous location where exception was thrown ---&#x000D;&#x000A;   at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)
```

General.ruleset
```diff
@@ -0,0 +1,110 @@
...
+<RuleSet Name="General C# rules" Description="" ToolsVersion="16.0">
```
Piedone: We don't need https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1204.md.

Piedone (changes requested):

General.ruleset
```diff
@@ -0,0 +1,110 @@
...
+  <Rules AnalyzerId="SonarAnalyzer.CSharp" RuleNamespace="SonarAnalyzer.CSharp">
+    <Rule Id="S101" Action="None" />
+    <Rule Id="S103" Action="Warning" />
+    <Rule Id="S104" Action="Warning" />
+    <Rule Id="S105" Action="Warning" />
+    <Rule Id="S1067" Action="Warning" />
+    <Rule Id="S107" Action="Warning" />
+    <Rule Id="S1109" Action="Warning" />
+    <Rule Id="S1128" Action="None" />
+    <Rule Id="S113" Action="Warning" />
```
Piedone, line 99: Do you have some personal experiences with this (https://rules.sonarsource.com/csharp/RSPEC-113), where does it actually make sense to add a new line at the end of the file? While git does tell if there's no new line, it's not an error or anything, just an indicator.

General.ruleset
```diff
@@ -0,0 +1,110 @@
...
+    <Rule Id="SA1507" Action="None" />
+    <Rule Id="SA1516" Action="None" />
+    <Rule Id="SA1519" Action="None" />
+    <Rule Id="SA1600" Action="None" />
+    <Rule Id="SA1601" Action="None" />
+    <Rule Id="SA1602" Action="None" />
+    <Rule Id="SA1633" Action="None" />
+    <Rule Id="SA1649" Action="None" />
+    <Rule Id="SX1101" Action="Warning" />
+    <Rule Id="SX1309" Action="Warning" />
```
Piedone: Is it possible to make exceptions on the field name? In Orchard we commonly have

```
private readonly IStringLocalizer T;
private readonly IHtmlLocalizer H;
```

fields.

General.ruleset
```diff
@@ -0,0 +1,110 @@
...
+    <Rule Id="CA1304" Action="None" />
+    <Rule Id="CA1305" Action="None" />
+    <Rule Id="CA1307" Action="None" />
+    <Rule Id="CA1308" Action="None" />
+    <Rule Id="CA1820" Action="None" />
+  </Rules>
+  <Rules AnalyzerId="Microsoft.NetCore.CSharp.Analyzers" RuleNamespace="Microsoft.NetCore.CSharp.Analyzers">
+    <Rule Id="CA1825" Action="None" />
+  </Rules>
+  <Rules AnalyzerId="SonarAnalyzer.CSharp" RuleNamespace="SonarAnalyzer.CSharp">
```
Piedone, line 89: Do you have some experience with this? https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1133.md I don't see why wouldn't we allow multi-attributes.

General.ruleset
Piedone: The default of this is 10: https://rules.sonarsource.com/csharp/RSPEC-1541 Is this what you'd recommend too? Same with https://rules.sonarsource.com/csharp/RSPEC-138 and 80 lines, https://rules.sonarsource.com/csharp/RSPEC-1067 and 3 operators. Can these be configured?

Piedone: This doesn't allow protected fields: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1051?view=vs-2019 While I agree on not having public non-`const` fields this seems too restrictive. If it can be modified to prevent public fields then great.

Piedone: I'm not sure about this: https://rules.sonarsource.com/csharp/RSPEC-927 Sometimes parameters in the base class coming from an external framework doesn't have the best name in the context we use them.

Piedone: This won't be needed, on the off chance we use regions they are actually useful: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1124.md

Piedone: Is there any configurability of this? https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md If not then please disable it.

Piedone: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1611.md would be useful but unfortunately, it also fires for the first parameters of extension methods that we never document as they're trivial. So unless we can exclude that case please disable it. https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1612.md has the same issue.

Piedone: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1615.md fires for trivial cases too, like when a method returns a `Task` or a `bool` (but we've already documented in the summary what the method does).

Piedone (changes requested):

General.ruleset
Piedone: What should we do if rules from multiple tools overlap? E.g. this one: https://rules.sonarsource.com/csharp/RSPEC-3963 comes from Sonar but also from .editorconfig so there's a warning even if it's suppressed for .editorconfig:

image

Should we just use Sonar?

.editorconfig
YevgeniyShunevych: Right.

General.ruleset
YevgeniyShunevych: In what case do you get this warning? When you open ruleset within Visual Studio with solution that has projects referencing used analyzers, everything should be OK. It required analyzer packages to be somewhere.

General.ruleset
YevgeniyShunevych: There is no possibility to add some exceptional cases for rules. Only suppress each warning individually or disable the rule completely.

But I would not use `T` and `H` for field names, besides of who recommends this unusual naming. Naming at least as `texts`/`strings`/`stringLocalizer` and `html`/`htmlLocalizer` doesn't look bad for me.

General.ruleset
YevgeniyShunevych: They are allowed.

But instead of:

```cs
[ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(MyCodeFixProvider)), Shared]
```

prefer:

```cs
[ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(MyCodeFixProvider))]
[Shared]
```

I agree with this rule. But if it breaks your code styles, we can disable this rule.

General.ruleset
YevgeniyShunevych: As for me, I'm completely agree with rule. This rule can reveal badly architectured code and human naming mistakes.

YevgeniyShunevych: No configuration here, as well. I know you don't follow this ordering system of class members, but I would consider it one more time, as this is the commonly used ordering.

YevgeniyShunevych: I would leave Sonar one. But you decide what to do with `.editorconfig` setting.

YevgeniyShunevych: Unfortunately, no configuration options. Either rule enabled or disabled. As for me, unless you develop games with C# and Unity 3D, absolutely all fields should be private. Decide.

.editorconfig
Piedone: OK then.

General.ruleset
Piedone: I tested it with the Finitive solution in this branch: https://github.com/Lombiq/Finitive/tree/issue/FINI-145 Note that for this to work I've already changed these (but didn't commit):
image

I get this when I open the solution and it never goes away.

General.ruleset
Piedone: I mean, it's possible in C# of course, but I don't see the rationale behind this rule. Why would you use the second version, why do you prefer it?

General.ruleset
Piedone: The thing is that these have been Orchard conventions for the past decade, and all code in Orchard itself, as well as in the Orchard ecosystem, uses them (so it would be strange for others to see something else in our code).

However, one workaround to still satisfy the private field with underscore rule would be to keep these localizers in `public IStringLocalizer T { get; }`. But I don't like the idea of introducing a property just for the sake of this.

Checking it out, the rule itself is not too complex: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/b6b9b0249a142132453873a677b7f09d7d65f34b/StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SX1309FieldNamesMustBeginWithUnderscore.cs So we could create our own analyzer package (see: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2265) and use a modified copy of this just to exclude these two items.

I realize this is out of scope of this project so I'll think about it.

General.ruleset
Piedone: We don't want to use two rules for the same thing so please disable this in .editorconfig with the comment that we're using RSPEC-3963 instead. In subsequent cases please do the same.

General.ruleset
Piedone: I've pushed this so I can work on custom analyzers in another branch.

General.ruleset
YevgeniyShunevych: When you read the sources you can visually miss some of attributes if there are several on the same line. So in reasons of code readability. If you want, we can disable this rule.

Piedone: I see the point. However, in practice we split large attribute lists already but not having this rule keeps it open to merge trivial ones (e.g. in UI tests where there are a lot of `[Theory, Chrome]` attributes that would be pointless to break into multiple lines). Readability is indeed important but in this case, I think the decision is not this simple, unfortunately. But the 200-character line warning takes care of a lot of such readability issues already (we use 120 characters as a soft limit, so it's perfect to have a much more intrusive warning above that).

So, let's disable it, please.

General.ruleset
YevgeniyShunevych: I haven't faced problems related to missing line break in the end of file. But I know that lot of people, as well as me, follow this "tradition".

General.ruleset
YevgeniyShunevych: No configuration here, as well. I do follow these rules mostly (exceptions happen). It pushes you to split complex methods into smaller ones, which makes everyone happier. If you want we can disable these rules.

YevgeniyShunevych: The rules are the rules. Do you want to disable this one?

YevgeniyShunevych: Disabled.

General.ruleset
YevgeniyShunevych: Disabled.

General.ruleset
YevgeniyShunevych: Disabled.

YevgeniyShunevych: Disabled.

YevgeniyShunevych: Disabled.

Piedone: Yes please.

Piedone: Let's disable it, please.

General.ruleset
```diff
@@ -0,0 +1,115 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Piedone, line 1: SA1408 and S1067 seem to be very similar. Which one would you recommend?

General.ruleset
```diff
@@ -0,0 +1,115 @@
...
+<RuleSet Name="General C# rules" ToolsVersion="16.0">
```
Piedone, line 2: Disable https://rules.sonarsource.com/csharp/RSPEC-3626 too, please. It suggested three items but all three were wrong and would break the code.

Piedone, line 2: Disable https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1118.md please.

Piedone, line 2: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1116.md and https://rules.sonarsource.com/csharp/RSPEC-1199 are the same. Which one do you recommend?

Piedone (changes requested):

Build.props
```diff
@@ -0,0 +1,29 @@
...
+  </ItemGroup>
+  
+  <Target Name="DisableAnalyzers" BeforeTargets="CoreCompile" Condition="'$(RunCodeAnalysis)' == 'false'"> 
+    <ItemGroup> 
+      <Analyzer Remove="@(Analyzer)" /> 
+    </ItemGroup> 
+  </Target>
+  
+  <Target Name="Copy editorconfig" AfterTargets="AfterResolveReferences">
+    <Copy SourceFiles="$(SolutionDir)Lombiq.Analyzers\.editorconfig" DestinationFolder="$(SolutionDir)" SkipUnchangedFiles="true" />
```
Piedone: When a project is built alone, without a solution, then `$SolutionDir` here will be undefined. We need to build projects individually sometimes in our CI, e.g. like this:

`dotnet build src\Utilities\Lombiq.Gulp.Extensions\Lombiq.Gulp.Extensions.csproj`

However, this failed with:

> Lombiq.Analyzers\Build.props(27,5): error MSB3030: Could not copy the file "*Undefined*Lombiq.Analyzers\.editorconfig" because it was not found.

Can we fix this? Or should we disable the analyzers in this case (something that if I'm not mistaken will work on later)?

Build.props
```diff
@@ -0,0 +1,29 @@
+<Project>
```
Piedone, line 1: Is it possible to configure for a solution that we get warnings on build only when we explicitly want it (like with an MSBuild flag we talked about), otherwise just when we open a file?

The reason I'm asking is that this analysis makes compilation a lot longer. This is understandable of course, and we actually need it now in the beginning, when fixing all the issues first. However, after that, when we've fixed all the existing issues, during development we only need warnings for the developer when they open the given file. That way they'll see the issues but when they build locally they won't get the performance penalty. But in CI we'd still do a full build, so if they didn't care about fixing something it'll still break there.

Is this possible? So get full solution analyses on build in CI but only file-based analysis during local development?

General.ruleset
Piedone: I think I get it:

- That I didn't always get results for a whole solution is simply because after a rebuild it takes minutes to generate the warnings. Note that I need to do a rebuild, just building (F6) an already built solution doesn't bring up all the warnings from the solution. So that's clear, nothing to do, though it's strange it needs a rebuild (and if I close and reopen the solution where I've done a rebuild and seen all the warnings previously, I need to rebuild again).
- I get the "An instance of analyzer..." warnings when opening the solution after a `git clean -fxd`. It goes away after fully building the solution, then reopening it. This is correlated with the `System.TypeLoadException` I mentioned above. However, then eventually (perhaps after reopening the solution a few times) I get the same warnings for the same (already fully built, previously full working) solution. And then I do a rebuild, get all the warnings but also get the `System.TypeLoadException` ones and thus IntelliSense suggestions are broken... So this is flaky.

.editorconfig
YevgeniyShunevych: Removed.

Readme.md
YevgeniyShunevych: Right. I updated the text.

General.ruleset
YevgeniyShunevych: Disabled.

YevgeniyShunevych: So disable it?

YevgeniyShunevych: Disabled.

Build.props
YevgeniyShunevych: To disable code analysis locally when using Visual Studio you can update `Directory.Build.props` file in the repostory root as I mentioned here: https://github.com/Lombiq/Yevgeniys-plate/issues/1#issuecomment-663124267

Build.props
YevgeniyShunevych: I've updated `Build.props` to use `$(MSBuildThisFileDirectory)` instead of `$(SolutionDir)`. It should fix a problem. Let me know if it doesn't.

General.ruleset
YevgeniyShunevych: Disabled.

YevgeniyShunevych: Disabled.

General.ruleset
YevgeniyShunevych: Disable this rule?

General.ruleset
YevgeniyShunevych: Disable this rule?

General.ruleset
YevgeniyShunevych: I played a bit with if statements to brake these rules. And these rules are not similar, they check different code issues. So I would leave both these rules enabled.

General.ruleset
YevgeniyShunevych: They are different. Check their descriptions one more time.

General.ruleset
Piedone: Yeah, I didn't update you on this, sorry. Let's keep it. While I don't see much of a point VS does the same by default and in the official source the convention is the same.

General.ruleset
Piedone: I'm still working on this and the other open questions to see what we want with the team, and how a custom analyzer is possible.

Build.props
Piedone: Looks good.

Build.props
Piedone: That's not what I'm talking about here. As I see it, analysis runs in two cases:
- When you open the file, just for that file.
- When you build the solution, for the whole solution.

The file-based one doesn't depend on the build because it also works when you open a freshly cloned solution, before building it.

So my question is, can we configure this behavior in a way that during local development (by default) we'd only see warnings for the currently opened files and not run analysis on build (just in CI)?

General.ruleset
Piedone: I don't know what's happening here. Opening the solution today doesn't show the `System.TypeLoadException` warnings, even though I didn't do anything between now and the warnings being there, apart from shutting down and starting my PC. Don't you experience the same? Like when you freshly clone the Finitive repo, checkout issue/FINI-145, and open the solution?

General.ruleset
Piedone: Hmmm indeed, these are completely different. I've mixed up something, nevermind.

General.ruleset
Piedone: Yeah, my mistake.

General.ruleset
```diff
@@ -0,0 +1,119 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Piedone, line 1: Isn't https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1401.md the same as CA1051 that you disabled?

Piedone, line 1: There's a curious thing:

This code trips [SA1116](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1116.md), rightly so:

```
public Task UnverifiedUserShouldBeRedirectedToEmailConfirmationPageAfterLogin(Browser browser) =>
    ExecuteTestAfterSetup(context =>
    {
        context.SignUp();

        context.LogOff();

        context.LogIn(OrdinaryUser.Email, OrdinaryUser.Password);
        AssertEmailConfirmationPage(context);
    }, browser);
```

See `EmailConfirmationTests` in the Finitve solution (or almost any other UI test class there).

However, this doesn't cause a warning:

```
public override IDisplayResult Edit(DocumentPart part, BuildPartEditorContext context) =>
    Initialize<EditDocumentPartViewModel>($"{nameof(DocumentPart)}_Edit", model =>
    {
        model.EndDateUtc = part.EndDateUtc;
        model.InternalNotes = part.InternalNotes;
        model.PrimaryContactEmail = part.PrimaryContactEmail;
        model.SecondaryContactEmail = part.SecondaryContactEmail;
        model.SignatureState = part.SignatureState;
        model.StartDateUtc = part.StartDateUtc;
        model.TermLengthMonths = part.TermLengthMonhts;
        model.TermType = part.TermType;
        model.ExternalPath = part.ExternalPath;
    }).Location("Content: 10");
```
See `DocumentPartDisplayDriver` (or a lot of other *Driver classes).

Why is this?

Build.props
YevgeniyShunevych: As far as I know there are only 2 options in Visual Studio which you can configure:

image

image

General.ruleset
YevgeniyShunevych: Strange. I have no idea. Nothing is perfect.

YevgeniyShunevych: I disabled SA1401 also then.

General.ruleset
YevgeniyShunevych: I don't remember either I had `System.TypeLoadException`.

_Like when you freshly clone the Finitive repo, checkout issue/FINI-145, and open the solution?_
Did it, built solution and everything is fine.

Piedone: Hmm, let me ask my colleagues, it may be due to the accented character in the path ("Zoltán") though it going away after a while suggests it's not because of that.

General.ruleset
Piedone: Well, I know :) but do you have any suggestions on how to catch the formatting issue in the second case too? Any other rules?

General.ruleset
Piedone: Are you on VS 16.6.5 Pro too?

Piedone: It turns out this was the Microsoft Code Analysis extension. I've added a note on this into the Readme.

Build.props
Piedone: Hmm, the "Background analysis scope" was already on "Open documents" for me. Doesn't seem to do much apparently.

The filter in the Error List window doesn't seem to do much for me, Build Only and IntelliSense Only shows exactly the same items.

So isn't there a way to disable analyzes during build?

Piedone: How about this? By default, the packages are removed on build (there are MSBuild steps running just before it) so it won't make the build slower. However, otherwise the packages are referenced so we get squiggly lines in the editor.

Then, you can disable this removal on build, i.e. enable code analysis on build, by either changing your solution's Directory.Build.props (for local development) or via a `dotnet build` command line parameter (for CI).

General.ruleset
Piedone: We decided to keep this, better to have consistency.

Piedone: Let's keep these. We've decided to refactor the code base gradually and these warnings might be helpful in that.

General.ruleset
YevgeniyShunevych: I don't know. You can, for example, raise an issue on StyleCop repository regarding that.

Piedone: Fair enough, will do.

Piedone: I see now that the issue is that that this is not actually covered by SA1116 but [SA1117](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1117.md). I enabled this by adding a `<Rule Id="SA1117" Action="Warning" />` line to the ruleset but it still doesn't notice `DocumentPartDisplayDriver`. I think it should provide a warning for both `Test` methods in the following classes but only works for the first one:

```
    public class Ok
    {
        public void Test() =>
            Invoked(input =>
            {
            }, "second argument");

        private void Invoked(Action<int> action, string secondParameter)
        {
        }
    }

    public class NotOk
    {
        public void Test() =>
            Invoked("first argument", input =>
            {
            });

        private void Invoked(string firstParameter, Action<int> action)
        {
        }
    }
```

Could you check please?

General.ruleset
```diff
@@ -0,0 +1,125 @@
...
```
Piedone, line 1: I get this when opening the file in VS:
image
Would you recommend moving these to the .editorconfig? Would they be executed during build still?

Build.props
Piedone, line 1: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1040?view=vs-2019 and https://rules.sonarsource.com/csharp/RSPEC-4023 are the same, please choose one to keep.

General.ruleset
Piedone: It seems that it's not possible to just include analyzers as code in the project and use them in the solution, you have to build a proper project (or actually, multiple) for them and package them up (and have the compiler SDK installed). So my initial idea of creating a copied analyzer with the localizers as exceptions won't work... I've disabled it then.

General.ruleset
Piedone: I've disabled this.

Piedone: There is configurability for this actually, see: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#element-order So please do configure it according to our documentation.

Build.props
Piedone, line 1: Please configure mandatory documentation but just for interfaces: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#documentation-requirements

General.ruleset
YevgeniyShunevych: I would not do that. This way you will lose an ability to manage ruleset visually with the VS editor, which is useful, as for me.

Piedone: Fair enough.

General.ruleset
YevgeniyShunevych: The same for me. I don't know why it doesn't generate a warning. But at least it works for the first case. As for me, we can live with this.

Build.props
YevgeniyShunevych: I think analyzers should be enabled by default. During local development people can make a change to `Build.props` file to disable analyzers temporarily for faster builds.

Piedone: I agree that people should see analyzer warnings locally for what they work on. The issue is that running analysis for the whole solution, even if they only changed two lines, on every build, makes the build 2x slower. This is a hit on productivity. Hence why my suggestion of the above: If I'm not mistaken that technique would provide the best of both worlds, allowing warnings locally for open files while not making build slower, and at the same time allow full analysis on CI builds where the slowdown doesn't matter.

General.ruleset
Piedone: I noticed that SA1117 works in some places so I've opened a StyleCop issue about it: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3183

Piedone: It turns out this is likely a StyleCop bug.

Piedone: If you happen to know an alternative rule from Sonar or FxCop (I checked but didn't find one) then please add that, otherwise, we'll leave this as such.

Build.props
YevgeniyShunevych: I don't completely get it. You can configure it as you want.

General.ruleset
YevgeniyShunevych: I also don't know another similar rule.

Build.props
YevgeniyShunevych: I've disabled Sonar's S4023 rule. As for me the priority of these 3 analyzer packages is:

1. FxCop (as it is developed by Microsoft)
1. StyleCop (because it is being useful for many years)
1. Sonar

So I would better disable the duplicate rule of analyzer with lower priority.

YevgeniyShunevych: Added `stylecop.json` file with needed configuration.

Piedone: Agree, great.

General.ruleset
Piedone: Let's leave this as such then.

Build.props
Piedone: Maybe I'm not clear enough but I'd prefer you do this since it'll require digging into the depths of MSBuild.

So now these analyzer packages are added, and thus when you open a file you get analysis for just that file (this happens without any build, even if you open a freshly cloned solution). And when you do a build you get analysis for the whole solution.

Both of these are great but this makes the build much slower, which impacts productivity since during local development you need to iterate quickly and can't wait 5 minutes for every build. You also don't need to run analysis for the whole solution every time. So we need some solution for a developer to be able to see analyzer warnings when writing code but not be slowed down by slow builds. In CI though we always want a full thorough analysis and make the build fail when there are any issues that the developer didn't address.

The question is, how to achieve this? We don't want devs to completely disable analysis locally as they have to see issues for the code they work on, we need IDE support. Hence my idea:

- In CI everything happens as it happens now, so full analysis on build.
- Locally. the analyzer packages are referenced when editing but removed just when build happens. There are steps in MSBuild to hook into this event but I don't know from the top of my head. If I'm correct this will let analyzers run during editing but effectively disable them during build.

As I said, I'm happy to expand the scope and budget of our project for this.

Piedone: This should show up under SA1600, right? This seems that it doesn't run on build, just when you open a given file.

YevgeniyShunevych: Right, SA1600. Try to rebuild solution. I've got 233 warnings of SA1600.

General.ruleset
YevgeniyShunevych: Cannot find your docs for this. Could you please compose an ordered list of these values: "kind", "accessibility", "constant", "static", "readonly" that you think suits your standards? I can then add this configuration to `stylecop.json`. I also guess that "kind" should be excluded, as you follow another (comparing to [SA1201](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md)) class members order.

Piedone: See [here](https://orcharddojo.net/orchard-resources/CoreLibrary/DevelopmentGuidelines/CodeStyling) under "Have a standard ordering of members..." For everything else not indicated there please just use the StyleCop default and we'll fine tune it later.

Build.props
Piedone: I tried that but they won't show up, not even if I switch the rule to Error. Does `dotnet build Finitive.sln --configuration Debug --no-incremental >build.log` output SA1600 violations for you? Not for me.

YevgeniyShunevych: Very strange. I also don't see SA1600 by using `dotnet build`. But I do have this warnings in Visual Studio.

YevgeniyShunevych: Actually, all SA16XX documentation rules were not present when using `dotnet build`. Adding `<GenerateDocumentationFile>true</GenerateDocumentationFile>` to `Build.props` seems that solved the problem. Please confirm that it is OK for you now.

General.ruleset
YevgeniyShunevych: OK, we have a problem here. Orchard standards look completely incompatible with the ones that StyleCop provides:

- `kind` (elements are ordered according to their kind (see [SA1201](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md) for this predefined order)) - Orchard class members order differs from the order defined in SA1201
- `accessibility` (elements are ordered according to their declared accessibility (see [SA1202](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1202.md) for this predefined order)) - valid for Orchard methods but not for fields: *Notice the order: static, private, protected, public, const, constructor, public, protected, private, static, inner classes.*
- `constant` (constant elements are ordered before non-constant elements) - in Orchard vice versa: *Constants just before the constructor* (means after all other fields).
- `static` (static elements are ordered before non-static elements) - in Orchard - static fields before and static methods after the rest.
- `readonly` (readonly elements are ordered before non-readonly elements) - in Orchard style I don't see anything about `readonly`.

So if you continue to follow Orchard's style, from this configuration only `readonly` probably can apply. And I don't see much advantage of this.

Build.props
Piedone: Works perfectly, thank you.

General.ruleset
Piedone: Oh damn.

I think the following ones would work though:

- accessibility
- constant (now that I see it we're actually doing what the rule says not what's documented)
- readonly

Please enable these and I'll check what happens, and adjust if necessary. We'd need [more configuration](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1610), unfortunately.

YevgeniyShunevych: Configured it this way. Please check it out.

Build.props
YevgeniyShunevych: I found useful project property `<RunAnalyzersDuringBuild>false</RunAnalyzersDuringBuild>` and added it to `Build.props`. Look like it does what you wanted. During solution build, analyzers are not run. But Visual Studio shows analyzer warnings for open documents. And it doesn't affect `dotnet build`, which continue to find all analyzer warnings.

Please check it out, @Piedone.

YevgeniyShunevych: Here is this property on project properties window by the way:

image

Build.props
```diff
@@ -0,0 +1,35 @@
...
+    <PackageReference Include="StyleCop.Analyzers" Version="1.1.118">
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+  </ItemGroup>
+  
+   <ItemGroup>
+    <AdditionalFiles Include="$(MSBuildThisFileDirectory)stylecop.json" Link="stylecop.json" />
+  </ItemGroup>
+  
+  <Target Name="DisableAnalyzers" BeforeTargets="CoreCompile" Condition="'$(RunCodeAnalysis)' == 'false'">
```
Piedone, line 27: Would you document how to use this with `dotnet build`, please?

Build.props
Piedone: Fantastic, that's just what we needed, thank you! Indeed a rebuild in VS now takes exactly the same amount of time than without the analyzers.

However, I noticed that the `dotnet build` output is not exactly the same. I've attached them [under the issue](https://github.com/Lombiq/Yevgeniys-plate/issues/1#issuecomment-674200033) because apparently that's not possible here. The "false" one is for `RunAnalyzersDuringBuild` = `false`.

I can't see any pattern apart from a bunch of warnings missing in the "false" file. I've built both with `dotnet build Finitive.sln --no-incremental > build.log` after each other ("false" first).

Piedone: Hmm, funnily enough, when executed in CI there are now more warnings then before.

And now if I run it again locally it's the same as with "true".

So it seems this was a red herring and works as expected!

General.ruleset
Piedone: Looks good. There are no SA1201 violations actually, just SA1202 ones. There are a lot of mismatches for accessibility though indeed, so that won't be usable. I've thus disabled that one.

Build.props
YevgeniyShunevych: Added "How to disable analyzers during dotnet build" section to Readme.md.

Piedone: Perfect.

PR #20, opened 2021-02-15
COLI-314: add a separate Orchard 1 ruleset to adjust independently of…

0liver:
… Orchard Core bases projects

### DIFF
--- a/Docs/AddingAnalyzers.md
+++ b/Docs/AddingAnalyzers.md
@@ -2,7 +2,7 @@
 
 
 
-## How to add the analyzers to your project repository
+## How to add the analyzers to your whole repository
 
 1. Add to *.gitmodules* file (we use the *tools* subfolder for the submodule's folder here but feel free to use something else):
    ```
@@ -12,7 +12,7 @@
        branch = dev
    ```
    *`path` can target anything but we suggest either the folder where the solution `.sln` file is located (mostly the repository root) or a "tools" subfolder therein.*
-2. Create a *Directory.Build.props* file in the folder where the solution *.sln* file is located (mostly the repository root) with the following content (if you've put the submodule in to a different folder then change the path):
+2. Create a *Directory.Build.props* file in the folder where the solution *.sln* file is located (mostly the repository root) with the following content (if you've put the submodule into a different folder then change the path):
    ```xml
    <Project>
      <Import Project="tools/Lombiq.Analyzers/Build.props"/>
@@ -31,6 +31,20 @@ This will use the analyzer configuration suitable for Orchard Core projects. If 
 </PropertyGroup>
 ```
 
+## How to add the analyzers to individual .NET Framework projects (not solutions)
+
+1. Same as above - add the .NET-Analyzers repository as a submodule to your repository.
+2. Create a *Directory.Build.props* file in every project folder you want to target, next to the *.csproj* file with the following content (import the *`NetFx.Build.props`* file instead of *Build.props*):
+   ```xml
+   <Project>
+     <Import Project="tools/Lombiq.Analyzers/NetFx.Build.props"/>
+   </Project>
+   ```
+3. The *NetFx.Build.props* will copy this project's *.editorconfig* file into every project folder that you've created a *Directory.Build.props* in, so you might want to gitignore those:
+    ```
+    .editorconfig
+    ```
+
 
 ## Introducing analyzers to an existing project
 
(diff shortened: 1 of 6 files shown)


Orchard1.ruleset
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Piedone, line 1: Add a quick note about this at the end of https://github.com/Lombiq/.NET-Analyzers/blob/dev/Docs/AddingAnalyzers.md#how-to-add-the-analyzers-to-your-project-repository as well, please, similar to how the note on General.ruleset is there.

Piedone (changes requested):

Orchard1.ruleset
0liver: I will update the documentation once done.

Docs/AddingAnalyzers.md
```diff
@@ -31,6 +31,22 @@ This will use the analyzer configuration suitable for Orchard Core projects. If
 </PropertyGroup>
 ```
 
+## How to add the analyzers to individual .NET Framework projects (not solutions)
+
+1. Same as above - add the .NET-Analyzers repository as a submodule to your repository.
+2. Create a *Directory.Build.props* file in every project folder you want to target, next to the *.csproj* file with the following content (import the *`NetFx.Build.props`* file instead of *Build.props*):
+   ```xml
+   <Project>
+     <Import Project="tools/Lombiq.Analyzers/NetFx.Build.props"/>
```
Piedone:
```suggestion
     <Import Project="../../../tools/Lombiq.Analyzers/NetFx.Build.props"/>
```

While it might not be exactly this it won't be just *tools* for sure.

BTW This is using forward slashes but the files in COLI contain "..\..\..\tools\Lombiq.Analyzers\NetFx.Build.props". Which one is good?

Docs/AddingAnalyzers.md
```diff
@@ -31,6 +31,22 @@ This will use the analyzer configuration suitable for Orchard Core projects. If
...
+   </Project>
+   ```
+3. The *NetFx.Build.props* will copy this project's *.editorconfig* file into every project folder that you've created a *Directory.Build.props* in, so you might want to gitignore those:
+    ```
+    .editorconfig
+    ```
+    
+This will use the analyzer configuration suitable for Orchard 1 projects. If you want to use the this in a non-Orchard .NET Framework app then use the *General.ruleset* file as described above.
```
Piedone:
```suggestion
This will use the analyzer configuration suitable for Orchard 1 projects. If you want to use this in a non-Orchard .NET Framework app then use the *General.ruleset* file as described above.
```

Docs/AddingAnalyzers.md
```diff
@@ -31,6 +31,22 @@ This will use the analyzer configuration suitable for Orchard Core projects. If
 </PropertyGroup>
 ```
 
+## How to add the analyzers to individual .NET Framework projects (not solutions)
```
Piedone: Add a note that this is only applicable for the legacy csprojs.

Docs/AddingAnalyzers.md
```diff
@@ -43,7 +59,7 @@ What to do if you're not starting a green-field project but want to add analyzer
...
-3. While this project can be used with .NET Framework, .NET Core and .NET 5 or later projects alike, it needs all projects to use the new SDK-style csproj format. You [can convert projects manually](https://docs.microsoft.com/en-us/dotnet/core/porting/#per-project-steps) or automatically with the [try-convert](https://github.com/dotnet/try-convert) utility. You'll also need to manually remove any leftover `packages.config` files and adjust `NuGet.config` files. 
+3. This project can be used with .NET Framework via *NetFx.Build.props* but the including project needs to use `PackageReference`s. You can convert your `packages.config` based project to `PackageReferences` using [this walkthrough](https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference). .NET Core and .NET 5 or later projects need to use the new SDK-style csproj format. You [can convert projects manually](https://docs.microsoft.com/en-us/dotnet/core/porting/#per-project-steps) or automatically with the [try-convert](https://github.com/dotnet/try-convert) utility. You'll also need to manually remove any leftover `packages.config` files and adjust `NuGet.config` files.
```
Piedone:
```suggestion
3. .NET compatibility:
    - .NET Core and .NET 5 or later projects can use this without anything special.
    - .NET Framework projects:
      - Projects using the new SDK-style csproj format don't need anything special. You [can convert projects manually](https://docs.microsoft.com/en-us/dotnet/core/porting/#per-project-steps) from the legacy csproj format or automatically with the [try-convert](https://github.com/dotnet/try-convert) utility. You'll also need to manually remove any leftover *packages.config* files and adjust *NuGet.config* files.
      - Projects using the legacy csproj format should use *NetFx.Build.props* as described above but the including project needs to use `PackageReference`s. You can convert your *packages.config*-based project to `PackageReference`s using [this walkthrough](https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference).  
```

Readme.md
```diff
@@ -4,7 +4,7 @@
...
+.NET code analyzers and code convention settings for [Lombiq](https://lombiq.com) projects, predominantly for [Orchard Core](https://www.orchardcore.net/) apps but also any .NET apps. We use these to enforce common standards across all our .NET projects, including e.g. all of our [open-source Orchard Core extensions](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions). If you contribute to our open-source projects while using that solution you'll be guided by these rules, too. You can check out a demo video of the project [here](https://www.youtube.com/watch?v=dtbGRi3Cezs). There is also support for .NET Framework projects, as long as they use *PackageReference* for their dependencies (in contrast to packages.config).
```
Piedone:
```suggestion
.NET code analyzers and code convention settings for [Lombiq](https://lombiq.com) projects, predominantly for [Orchard Core](https://www.orchardcore.net/) apps but also any .NET apps. We use these to enforce common standards across all our .NET projects, including e.g. all of our [open-source Orchard Core extensions](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions). If you contribute to our open-source projects while using that solution you'll be guided by these rules, too. You can check out a demo video of the project [here](https://www.youtube.com/watch?v=dtbGRi3Cezs). There is also support for .NET Framework projects, as long as they use `PackageReference` for their dependencies (in contrast to *packages.config*).
```

Also add a note about csprojs being the legacy version here.

NetFx.Build.props
```diff
@@ -0,0 +1,60 @@
...
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.321">
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <AdditionalFiles Include="$(MSBuildThisFileDirectory)stylecop.json" Link="stylecop.json" />
```
Piedone, line 48: Hmm, does this work? The files don't show up under the projects.

Piedone (changes requested):

NetFx.Build.props
0liver: I don't think they should since they are not being copied. They are just referenced from the project - and their path is given in the `Include` property.

Docs/AddingAnalyzers.md
0liver: Both are OK - here I used the same syntax as further up in the same file.

Do you think it's necessary to adjust the files in COLI with regard to this documentation?

Piedone: No, just the doc is OK.

NetFx.Build.props
Piedone: They do show up for SDK projects like this:
image

It might mean they don't get applied here. Please check if the config in stylecop.json actually get applied it's just not showing or is it indeed missing.

Piedone: It actually works, nothing to do here then.

Piedone (approved):

Orchard1.ruleset
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Orchard 1 C# rules" ToolsVersion="16.0">
```
Piedone, line 2: This can be probably disabled, since Orchard validates all requests by default? https://rules.sonarsource.com/csharp/RSPEC-4564 As well as SCS0016. "SCS0017 C# Request validation is disabled" can remain as it's good to double-check.

Piedone (changes requested):

Orchard1.ruleset
Piedone: I was impatient and implemented this :).

PR #38, opened 2022-01-24
Revert "ADHOC-166: Upgrading package versions and LangVersion"

Piedone:
Reverts Lombiq/.NET-Analyzers#36


Piedone:
Merged the PR by mistake.

PR #39, opened 2022-01-27
INFRA-92: NuGet publish

Piedone:


Docs/AddingAnalyzers.md
```diff
@@ -45,6 +45,11 @@ For at least Visual Studio and JetBrains Rider you don't need any further setup
 ```
 
 
+## How to add the analyzers to one SDK-style project from NuGet
+
+The recommended approach for SDK-style projects is adding .NET Analyzers as a submodule as explained above due to the increased control you have over configuration and because that way it can cover a whole solution, not just a single project. However, you can also use the [NuGet package](https://www.nuget.org/packages/Lombiq.Analyzers/) to install it for just one project. Once you add the package to your project, all analyzers will be applied.
```
barthamark: Can't you just add it to a `Directory.Build.targets` in the project (not here) so that it affects all projects? It's instead of adding it individually. It won't be self-contained, though.

Piedone: Good point, added it.

PR #53, opened 2022-04-26
OCC-3: Fix linux-specific problems in Lombiq.Analyzers.

sarahelsaig:
[OCC-3](https://lombiq.atlassian.net/browse/COMM-3)

PR #55, opened 2022-05-12
OSOE-49: Enforce Windows newlines for C# files to avoid false positives with IDE0055 warning.

sarahelsaig:
OSOE-49

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
 * text=auto
 
-# Enforce Windows newlines for C# files to avoid false positives with IDE55 warning.
+# Enforce Windows newlines for C# files to avoid false positives with IDE0055 warning.
 *.cs text eol=crlf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 * text=auto
 
 # Enforce Windows newlines for C# files to avoid false positives with IDE0055 warning.
-# See https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/106
+# See https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/106 for more information.
 *.cs text eol=crlf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 * text=auto
 
 # Enforce Windows newlines for C# files to avoid false positives with IDE0055 warning.
+# See https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/106
 *.cs text eol=crlf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,4 @@
 * text=auto
+
+# Enforce Windows newlines for C# files to avoid false positives with IDE55 warning.
+*.cs text eol=crlf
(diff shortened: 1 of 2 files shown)


Piedone (approved):

Piedone (changes requested):
With the current SDK, [OCC-3: Fix linux-specific problems in Lombiq.Analyzers. by DAud-IcI · Pull Request #53 · Lombiq/.NET-Analyzers](https://github.com/Lombiq/.NET-Analyzers/pull/53/files#diff-2dde510688211f1479ac8bbd63641430561100e5ac6b5e50f6e8821a9322bfb9)  shouldn’t be needed, this is why I removed it under [Removing now unneeded Microsoft.Net.Compilers.Toolset reference · Lombiq/.NET-Analyzers@52016af](https://github.com/Lombiq/.NET-Analyzers/commit/52016af07c3ce298ccc2d3a2a26725f14457a746).

sarahelsaig:
That wasn't my experience, that's why I have added it back. What do you mean by "current" exactly?

Piedone:
\>=6.0.202, see https://github.com/dotnet/sdk/issues/23972

sarahelsaig:
That explains it, I had "6.0.102" installed on my Linux desktop. But if we have a specific minimum version requirement within the same major version that really should be documented.

Piedone:
I'd give us (me) a pass here, since it was a workaround for a .NET SDK bug they introduced with a patch version :D.

sarahelsaig:
Anyway, we shouldn't remove the fix at least until May 22, because [6.0.200's lifecycle is still active](https://docs.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs#:~:text=Jul%20'23-,6.0.200,-17.1).

Piedone:
In principle I agree, I'd just really dread to re-add this workaround for these 9 days and have to keep in mind to remove it and retest everything.

sarahelsaig:
Even after may 22, it's not like everyone will instantly and magically switch over to the latest SDK. Why are you so eager to break backwards compatibility? Does this fix have any serious drawbacks?

Piedone:
Yeah, but we can still say we don't support older SDKs.

This is a risky workaround because it makes us use the compiler from NuGet instead of the SDK, which is unexpected and discouraged in [the package Readme](https://www.nuget.org/packages/Microsoft.Net.Compilers.Toolset/) too. It's a hack that we shouldn't keep until absolutely necessary, hence me also jumping on removing it the first time this was possible.

sarahelsaig:
> Yeah, but we can still say we don't support older SDKs.

I suppose we can say that. (I'm not happy about it tho.) Yet the main problem is that we didn't say anything, which confused the person I least want to see confused: myself. :D  Instead of deleting the code, I'm going to comment it out and add some text indicating that  SDK versions prior to 6.0.102 have to uncomment it.

Piedone:
I'm sorry :(. Sure! prior to 202, though.

Piedone (approved):

Piedone (approved):

0liver (approved):

PR #57, opened 2022-05-16
OSOE-110: Fix that the NuGet version of Lombiq Analyzer doesn't detect any analyzer violations

sarahelsaig:
OSOE-110

## Release Notes

- NuGet package dependencies are handled differently for the submodule and the NuGet package (via _nupkg_ file instead of `<PackageReference>`) but both are still derived from the _CommonPackages.props_ file.


TestSolutions/Lombiq.Analyzers.PackageReference/Lombiq.Analyzers.PackageReference.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="..\AnalyzerViolations.cs">
+      <Link>AnalyzerViolations.cs</Link>
+    </Compile>
+  </ItemGroup>
+
```
0liver: Looks like some accidental whitespace made its way into this line:

```suggestion

```

My suggestion is to remove the whitespace, not the whole line, but looks like GitHub doesn't understand that.

Lombiq.Analyzers/CommonPackages.props
```diff
@@ -0,0 +1,25 @@
...
+  <ItemGroup>
+    <AnalyzerPackage Include="AsyncFixer" Version="1.6.0"/>
+    <AnalyzerPackage Include="DotNetAnalyzers.DocumentationAnalyzers" Version="1.0.0-beta.59"/>
+    <AnalyzerPackage Include="Meziantou.Analyzer" Version="1.0.701"/>
+    <AnalyzerPackage Include="Microsoft.CodeAnalysis.CSharp.CodeStyle" Version="4.1.0"/>
+    <AnalyzerPackage Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0"/>
+    <AnalyzerPackage Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.2.32"/>
+    <AnalyzerPackage Include="SecurityCodeScan.VS2019" Version="5.6.2"/>
+    <AnalyzerPackage Include="SonarAnalyzer.CSharp" Version="8.38.0.46746"/>
+    <AnalyzerPackage Include="StyleCop.Analyzers" Version="1.2.0-beta.406"/>
```
0liver, line 15: How is this the same as before?

Lombiq.Analyzers/CommonPackages.props
```diff
@@ -0,0 +1,25 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="@(AnalyzerPackage)" Version="%(Version)" IncludeInPackage="true">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers;</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
```
0liver: Very slick. DRY at its best.

TestSolutions/AnalyzerViolations.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Lombiq.Analyzers.ProjectReference;
+
+// This file has intentional analyzer violations. Only edit the one in the TestSolutions directory. Any copy in the
+// subdirectories will be copied over during build. Copying is necessary so the source files are in a directory with an
+// existing parent .editorconfig file. (Linking will cause them to physically be outside of the .editorconfig's area of
+// effect.) Also please commit the updated version everywhere, because if the file is not present before the build
+// starts then MSBuild won't include it despite the <Copy> task. For this purpose the <Copy> task is only there to
```
0liver, line 7: > if the file is not present before the build starts then MSBuild won't include it

I've had to work around this behavior in NE, too, and was able to get it to work by adding the generated files to the `EmbeddedResource` item group like [this](https://github.com/Lombiq/NodeJs-Extensions/blob/dev/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets#L124).

sarahelsaig: Great, thanks Oliver, you are a true life saver. Because after this the _.editorconfig_ (which is copied by the _Lombiq.Analyzers/Build.props_) started to do just the same. (see [this](https://github.com/Lombiq/.NET-Analyzers/runs/7567993084?check_suite_focus=true); the test for "Local ProjectReference" passes because I committed the .editorconfig [here](https://github.com/Lombiq/.NET-Analyzers/pull/57/commits/1d4231838fb4d97170fc316e59f96a090645653b))

Lombiq.Analyzers/Build.props
```diff
@@ -63,7 +67,20 @@
...
-  <Target Name="Copy editorconfig" AfterTargets="AfterResolveReferences" Condition="'$(SolutionDir)' != '*Undefined*'">
+  <PropertyGroup Condition="'$(NuGetBuild)' == 'true'">
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+    <MSBuildTreatWarningsAsErrors>true</MSBuildTreatWarningsAsErrors>
+    <CodeAnalysisTreatWarningsAsErrors>True</CodeAnalysisTreatWarningsAsErrors>
+  </PropertyGroup>
+
+  <Target Name="Copy editorconfig"
+          AfterTargets="AfterResolveReferences"
+          BeforeTargets="EmbeddModuleAssets;OrchardCoreEmbedModuleAssets;Compile"
```
0liver: You'll probably only need `Compile` here - the other two targets are Ochard-specific.

TestSolutions/AnalyzerViolations.cs
0liver: Let's see!

sarahelsaig: [Sadly it doesn't work.](https://github.com/Lombiq/.NET-Analyzers/runs/7568514194?check_suite_focus=true)

0liver: Can't you copy `AnalyzerViolations.cs` at an even earlier stage? Like `BeforeBuild` e.g.?

0liver: Actually, [even earlier](https://stackoverflow.com/a/51609999/177710): `PrepareForBuild`.

sarahelsaig: Already tried it, no change. I'm currently using ResolveProjectReferences which is even earlier; it's the earliest possible target (I've checked it in the build logs).
Out of desperation I've even tried using Clean. Then if I do `dotnet clean; dotnet build` it works, but `dotnet build --no-incremental` (rebuild) doesn't even though that does include the Clean step. And just to spite me, `dotnet build --no-incremental; dotnet build --no-incremental` still works. So it really comes down to MSBuild not (always) adding files created during execution.

0liver: For me the order is the opposite:

image

sarahelsaig: Weird. But I won't hang up on it since it doesn't work either way

.gitignore
```diff
@@ -8,3 +8,8 @@ wwwroot/
 node_modules/
 *.user
 .pnpm-debug.log
+/.editorconfig
+/TestSolutions/Lombiq.Analyzers.*/.editorconfig
```
0liver: You can generalize this to:
```suggestion
/TestSolutions/**/.editorconfig
```

.gitignore
```diff
@@ -8,3 +8,8 @@ wwwroot/
...
+/TestSolutions/Lombiq.Analyzers.*/AnalyzerViolations.cs
+/TestSolutions/Lombiq.Analyzers.PackageReference/IndirectPackageReference/AnalyzerViolations.cs
```
0liver: Combine these two lines to:
```suggestion
/TestSolutions/Lombiq.Analyzers.*/**/AnalyzerViolations.cs
```

ConvertTo-Nuspec.ps1
```diff
@@ -0,0 +1,24 @@
+﻿param($Version)
+
+$projectPath = Join-Path $PWD Lombiq.Analyzers
+function Open-Xml([string]$File) { [xml](Get-Content (Join-Path $projectPath $File)) }
```
0liver: Looks like `Get-Xml` to me.

Invoke-BeforePack.ps1
```diff
@@ -0,0 +1,19 @@
+# This script tries to work around NuGet's validation and the NU5100 warning caused by the presence of these files
```
0liver:
```suggestion
# This script tries to work around NuGet's validation and the NU5100 warning caused by the presence of these files:
```

0liver (changes requested):

.gitignore
DAud-IcI: I used this to be similar as the line below, where it actually matters. (because `/**/` applies to `.` as well)

ConvertTo-Nuspec.ps1
sarahelsaig: Actually it should be `Read-Xml`. See [Get vs Read](https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.2#get-vs-read).

ConvertTo-Nuspec.ps1
```diff
@@ -0,0 +1,24 @@
+﻿param($Version)
```
0liver, line 1: We only execute this script for the `Lombiq.Analyzers` project, right?

Having the [extra build step](https://github.com/Lombiq/GitHub-Actions/pull/21/files#diff-5faf43e0bf23a1323690a2c9fb6921bd066a6ae504d8a982fb85ad25bc99e3e6R58) in the publish nuget GitHub action, which is a new extensibility point in the publish workflow, seems a but much.

While I understand your frustration with the whole issue and the many hacks you've had to build to achieve the goal, it seems to me that this should be only part of the Analyzers project, and not our general build .NET pipeline.

Couldn't this logic be moved into an MSBuild Target inside the project file and trigger it when `NuGetBuild=true`?

Invoke-BeforePack.ps1
```diff
@@ -0,0 +1,19 @@
+# This script tries to work around NuGet's validation and the NU5100 warning caused by the presence of these files:
+# - bin/Release/netstandard2.0/Lombiq.Analyzers.dll
+# - obj/Release/netstandard2.0/Lombiq.Analyzers.dll
+# The cause is probably that we don't actually do anything in the corresponding csproj, it only exists because we need
+# an entry point for `dotnet pack` (unlike the older `nuget pack` tooling, which worked with just the .nuspec alone).
+
+$release = Get-ChildItem Lombiq.Analyzers/bin/Release -Recurse -Include Lombiq.Analyzers.dll
+$releasePdb = $release.FullName -replace 'dll$','pdb'
+New-Item -Type Directory -Force lib
+Copy-Item $release lib
```
0liver: Copying the dll into the lib folder should fix the [NU5100](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5100) warning, no?

Invoke-BeforePack.ps1
```diff
@@ -0,0 +1,19 @@
...
+
+if (Test-Path $releasePdb)
+{
+    Copy-Item $releasePdb lib
+}
+else
+{
+    Write-Warning "There is no $releasePdb file!"
```
0liver: This will fail the build. Can't we just call `Copy-Item` for the pdb file, too, and then the build will automatically fail in cas of an unsuccessful copy task?

Lombiq.Analyzers.sln
```diff
@@ -0,0 +1,22 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30114.105
```
0liver: How come this uses VS 16? _Lombiq.OSOCE.sln_ contains these lines:
```suggestion
# Visual Studio Version 17
VisualStudioVersion = 17.0.32112.339
```

Lombiq.Analyzers.sln
```diff
@@ -0,0 +1,22 @@
+﻿
```
0liver, line 1: What exactly do we need this solution file for? We don't call `dotnet pack` on it, we can also build the project independently, ...

Lombiq.Analyzers/Build.props
```diff
@@ -24,6 +22,12 @@
     <_SkipUpgradeNetAnalyzersNuGetWarning>true</_SkipUpgradeNetAnalyzersNuGetWarning>
   </PropertyGroup>
 
+  <!-- Don't run analyzers during Visual Studio build, not to slow down development. They still run during dotnet build.
+       By making this conditional, you can still set the variable to true even when using the NuGet package. -->
+  <PropertyGroup Condition="'$(RunAnalyzersDuringBuild)' != 'true'">
+    <RunAnalyzersDuringBuild>false</RunAnalyzersDuringBuild>
+  </PropertyGroup>
```
0liver: I'd move the `Condition` onto the `RunAnalyzersDuringBuild` property and move the latter back up into the `PropertyGroup`.

Lombiq.Analyzers/Build.props
```diff
@@ -63,7 +67,24 @@
...
+    <CodeAnalysisTreatWarningsAsErrors>True</CodeAnalysisTreatWarningsAsErrors>
+  </PropertyGroup>
+
+  <Target Name="CopyEditorconfig" DependsOnTargets="LombiqNetAnalyzers" AfterTargets="AfterResolveReferences">
+  </Target>
+
+  <!-- This target should be used in CI before build, e.g. `dotnet msbuild -target:LombiqNetAnalyzers | Out-Null`. Use
+       `| Out-Null` in PowerShell or `>/dev/null 2>&1` in bash to suppress output because it will complain if not every
+       project has this target, but that's not really relevant. If you are using NuGet, then the argument should be
+       "-target:Restore;LombiqNetAnalyzers" instead (include the quotes) to load the pacakge first.  -->
```
0liver: [Currently](https://github.com/Lombiq/GitHub-Actions/pull/21/files#diff-4cf358b6f80e00ce9cd739a04c94dfad647b09ca5d6a524757944a1c611d6848R54), we call `dotnet msbuild '-target:Restore;LombiqNetAnalyzers' $Solution | Out-Null || bash -c 'true'`, so let's communicate the same here.

The last sentence is somewhat unclear, given that we already the what you suggest there, and also
> to load the pacakge first
probably means to restore all package dependencies first, right?

Lombiq.Analyzers/Build.props
```diff
@@ -63,7 +67,24 @@
...
-  <Target Name="Copy editorconfig" AfterTargets="AfterResolveReferences" Condition="'$(SolutionDir)' != '*Undefined*'">
-    <Copy SourceFiles="$(MSBuildThisFileDirectory).editorconfig" DestinationFolder="$(SolutionDir)" SkipUnchangedFiles="true" />
+  <PropertyGroup Condition="'$(NuGetBuild)' == 'true'">
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+    <MSBuildTreatWarningsAsErrors>true</MSBuildTreatWarningsAsErrors>
+    <CodeAnalysisTreatWarningsAsErrors>True</CodeAnalysisTreatWarningsAsErrors>
+  </PropertyGroup>
+
+  <Target Name="CopyEditorconfig" DependsOnTargets="LombiqNetAnalyzers" AfterTargets="AfterResolveReferences">
+  </Target>
```
0liver: Can I go home?

Lombiq.Analyzers/Build.props
```diff
@@ -63,7 +67,24 @@
     </ItemGroup>
   </Target>
 
-  <Target Name="Copy editorconfig" AfterTargets="AfterResolveReferences" Condition="'$(SolutionDir)' != '*Undefined*'">
-    <Copy SourceFiles="$(MSBuildThisFileDirectory).editorconfig" DestinationFolder="$(SolutionDir)" SkipUnchangedFiles="true" />
```
0liver, line 67: This should be restored, right, to keep the current behavior when using Analyzers as a submodule?

Lombiq.Analyzers/build/Lombiq.Analyzers.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <PropertyGroup>
+    <NuGetBuild>true</NuGetBuild>
+    <RestoreUseStaticGraphEvaluation>true</RestoreUseStaticGraphEvaluation>
```
0liver: We should probably try this out on all of OSOCE. I also like adding new things to my issues, but I remember that time when you snuck in file-scoped namespaces - Zoltán was very quick in making that a global thing. Maybe create an issue to adapt that everywhere, or at least discuss it in the open?

Here's how we could approach it: https://github.com/NuGet/docs.microsoft.com-nuget/issues/1941#issuecomment-717477626

TestSolutions/Directory.Build.props
```diff
@@ -0,0 +1,13 @@
+<Project>
+  <PropertyGroup>
+    <!-- You can enable analyzers for every build by adding this property. It works when you import Lombiq.Analyzers
+         from a NuGet package too.
+    <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild> -->
```
0liver: The same property is actually used below. Please merge the comment with that.

TestSolutions/Directory.Build.props
```diff
@@ -0,0 +1,13 @@
...
+
+    <!-- This property prevents the Directory.Build.props in the repository root. -->
+    <SkipRootBuildProps>true</SkipRootBuildProps>
```
0liver: This doesn't seem to actually be needed: https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022#search-scope. Just having the file here is what prevents the root _D.B.props_ from being imported.

TestSolutions/Lombiq.Analyzers.ProjectReference/Lombiq.Analyzers.ProjectReference.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
```
0liver: This property is already set in the [_Directory.Build.props_](https://github.com/Lombiq/.NET-Analyzers/pull/57/files#diff-01efe6d02cee6ef2712e9dca8109bb0c51439844179822ea35627b8bd1f1bc19R11).

TestSolutions/Lombiq.Analyzers.PackageReference/IndirectPackageReference/IndirectPackageReference.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+  </PropertyGroup>
+
+  <!-- This project demonstrates that the package reference doesn't transfer the analyzer effect to its dependents. If
+       you set ImportPackageAgain to "true" and run static code analysis on this soltuion you will get twice as many
+       warnings. This is the desired behavior, shows that the analyzers only activate when they are explicitly imported.
```
0liver, line 9: I seem to be missing something. I don't see any indirect reference to the Analyzers here.

TestSolutions/Lombiq.Analyzers.PackageReference/IndirectPackageReference/IndirectPackageReference.csproj
```diff
@@ -0,0 +1,24 @@
...
+  <ItemGroup Condition="'$(ImportPackageAgain)' == 'true'">
+    <PackageReference Include="Lombiq.Analyzers" Version="2.1.1-osoe-110.14" />
+  </ItemGroup>
+  
+  <!-- Dummy target for when ImportPackageAgain is disabled. -->
+  <Target Name="LombiqNetAnalyzers" Condition="'$(ImportPackageAgain)' != 'true'">
+  </Target>
+
+  <Target Name="CopyAnalyzerViolations" DependsOnTargets="LombiqNetAnalyzers" AfterTargets="LombiqNetAnalyzers">
+    <Copy SourceFiles="..\..\AnalyzerViolations.cs" DestinationFolder="$(MSBuildThisFileDirectory)" />
```
0liver, line 29: When I open the `Lombiq.Analyzers.PackageReference` solution in VS, the `AnalyzerViolations.cs` file is not copied into this project. It looks like I have to call the `LombiqNetAnalyzers` target to make that happen.

Is all of this only there for continuous integration tests? Where does this actually come into play? Is it [this line](https://github.com/Lombiq/GitHub-Actions/pull/21/files#diff-4cf358b6f80e00ce9cd739a04c94dfad647b09ca5d6a524757944a1c611d6848R54) that triggers the copying and then the solution build won't show any errors on the _AnalyzerViolations.cs_ file inside this project? But if there's no (indirect) reference to `Analyzers`, then that seems pointless. What am I missing?

TestSolutions/Lombiq.Analyzers.PackageReference/IndirectPackageReference/IndirectPackageReference.csproj
```diff
@@ -0,0 +1,24 @@
...
+
+  <ItemGroup Condition="'$(ImportPackageAgain)' == 'true'">
+    <PackageReference Include="Lombiq.Analyzers" Version="2.1.1-osoe-110.14" />
+  </ItemGroup>
+  
+  <!-- Dummy target for when ImportPackageAgain is disabled. -->
+  <Target Name="LombiqNetAnalyzers" Condition="'$(ImportPackageAgain)' != 'true'">
+  </Target>
+
+  <Target Name="CopyAnalyzerViolations" DependsOnTargets="LombiqNetAnalyzers" AfterTargets="LombiqNetAnalyzers">
```
0liver: From all I understand it seems to me that `DependsOnTargets="LombiqNetAnalyzers"` can be removed everywhere. Having that means that when building `CopyAnalyzerViolations`, MSBuild should first build `LombiqNetAnalyzers`, but you never invoke `CopyAnalyzerViolations` itself directly. Instead, you invoke `LombiqNetAnalyzers` from the build script which then triggers the build of `CopyAnalyzerViolations` thanks to the `AfterTargets="LombiqNetAnalyzers"` property.

Lombiq.Analyzers/Build.props
```diff
@@ -24,6 +22,12 @@
     <_SkipUpgradeNetAnalyzersNuGetWarning>true</_SkipUpgradeNetAnalyzersNuGetWarning>
   </PropertyGroup>
 
+  <!-- Don't run analyzers during Visual Studio build, not to slow down development. They still run during dotnet build.
+       By making this conditional, you can still set the variable to true even when using the NuGet package. -->
+  <PropertyGroup Condition="'$(RunAnalyzersDuringBuild)' != 'true'">
+    <RunAnalyzersDuringBuild>false</RunAnalyzersDuringBuild>
```
0liver: > By making this conditional, you can still set the variable to true even when using the NuGet package.

Actually, it's different than that. When you reference the NuGet package, the associated _.props_ file is imported from the consuming project before the beginning of its project file, so the condition here will never see the `RunAnalyzersDuringBuild` property being set in the consumer's project file.

But setting it from the _Directory.Build.props_ actually precedes the import of the _.props_ file from the NuGet package, and to respect that, your workaround is actually necessary. 😮

TestSolutions/Lombiq.Analyzers.ProjectReference/Lombiq.Analyzers.ProjectReference.csproj
```diff
@@ -0,0 +1,14 @@
...
+  </PropertyGroup>
+
+  <Target Name="CopyAnalyzerViolations" DependsOnTargets="LombiqNetAnalyzers" AfterTargets="LombiqNetAnalyzers">
+    <Copy SourceFiles="..\AnalyzerViolations.cs" DestinationFolder="$(MSBuildThisFileDirectory)" />
+  </Target>
+
+  <Import Project="..\..\Lombiq.Analyzers\Build.props" />
```
0liver: Move the `Import` of _.props_ files to the beginning of the project file in conformance with what happens when you reference the NuGet package. See my comment on _Build.props_ for more details.

0liver (changes requested):

ConvertTo-Nuspec.ps1
sarahelsaig: An extra build step costs effectively nothing. This way it's a formal extension point (that we can reuse if we ever need some of the more advanced nuspec features that `dotnet pack` doesn't provide) in the action instead of something custom-made and tacked-on. To my eyes, this makes it less hacky.

Invoke-BeforePack.ps1
sarahelsaig: It should, but it doesn't.

Invoke-BeforePack.ps1
sarahelsaig: Why would it fail the build? This is just a PS warning. afaik those don't get transformed into errors.

Lombiq.Analyzers.sln
sarahelsaig: That's a great question. :)

Lombiq.Analyzers.sln
sarahelsaig: We still use the solution file to decide which project files should be included in the pack.

Lombiq.Analyzers/Build.props
sarahelsaig: image

Lombiq.Analyzers/Build.props
sarahelsaig: That specific call is applicable for GHA, but I guess most Windows users don't have/know bash. Also it's not valid in Windows PowerShell either because the `||` operator is not supported there.

I replaced the "to load the pacakge first" with "to restore all package dependencies first".

Lombiq.Analyzers/Build.props
sarahelsaig: There is no escape, 😈 Also yes, this is actually necessary, because this way you only have to know about one target externally, the LombiqNetAnalyzers.

Lombiq.Analyzers/Build.props
sarahelsaig: The copy has been moved near the end, the current behavior for submodules is intact.

Lombiq.Analyzers/build/Lombiq.Analyzers.props
sarahelsaig: image
Ah Microsoft, never change.

ConvertTo-Nuspec.ps1
0liver: Making this a public extension point feels like YAGNI to me, when the current use-case is very specialized, for a single purpose, nothing we would ever reuse.

On the other hand, maybe that's just how extensible software is built. And that's generally a good thing. Time will tell.

Invoke-BeforePack.ps1
0liver: My fault. I was thinking of all the dotnet builds, but this is not part of them. OK then.

Lombiq.Analyzers/Build.props
0liver: Hm, works in NE: https://github.com/Lombiq/NodeJs-Extensions/blob/dev/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.props#L21 😕 

Is that Rider or Code?

Lombiq.Analyzers/Build.props
0liver: OK, ignore the `|| bash -c 'true'` part - I was trying to point to the fact that we use `-target:Restore;LombiqNetAnalyzers` ourselves, in contrast to the suggestion in the first line which is `-target:LombiqNetAnalyzers`. In that case the extra mention for the use of `nuget` (?) is just about the quotes `"`?

Lombiq.Analyzers/Build.props
sarahelsaig: Rider. It does build, but I don't want unnecessary redness.

Lombiq.Analyzers/Build.props
sarahelsaig: It's for nuget.

Lombiq.Analyzers/Build.props
0liver: Oh, wait, now I see that you're calling `LombiqNetAnalyzers` as a dependency of this empty target which will or will not do something depending on the type of reference we use. Devilishly clever indeed!

Add a comment inside the target please for those mortal humans who will come across this **MSBuild black magic**™ in the future.

Lombiq.Analyzers/Build.props
0liver: I see now - this target is basically split in two!

Lombiq.Analyzers/build/Lombiq.Analyzers.props
sarahelsaig: I guess I don't need this after all so I've removed it. We might want to add it to OSOCE as you suggested, but that can be a separate discussion.

TestSolutions/Lombiq.Analyzers.PackageReference/IndirectPackageReference/IndirectPackageReference.csproj
sarahelsaig: Indirect package reference means that this project references Lombiq.Analyzers.PackageReference which in turn references Lombiq.Analyzers. In other words we show that just because you import a project that uses Lombiq.Analyzers your project won't get "infected" by Lombiq.Analyzers against your will. Only directly importing Lombiq.Analyzers activates its effects.

TestSolutions/Lombiq.Analyzers.PackageReference/IndirectPackageReference/IndirectPackageReference.csproj
DAud-IcI: I can't reproduce that. When I open the solution in VS it immediately create copies the files before build so you don't have to do anything special. Similar for Rider too. I guess this is due to incremental compilation during build or something along those lines.

https://user-images.githubusercontent.com/4007293/182661120-131c33f3-c7f1-47b0-9a4a-bc0d8ba4e51a.mp4

The separate `LombiqNetAnalyzers` target is only for CI, so you don't have to build twice.

sarahelsaig: I missed to include the `AfterTargets="AfterResolveReferences"` here as well to splice it into the normal build process like it is in Lombiq.Analyzers/Build.props.

Invoke-BeforePack.ps1
0liver: If it doesn't fix it, can't we simply drop it?

Or fix the warning by following the [Solution](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5100#solution) paragraph, i.e.:
> ensure that all assembly files are placed inside a **framework-specific** folder under a lib folder

?

Lombiq.Analyzers/Build.props
0liver: I understand. Would it be too much to ask to file a bug with them? The extra `PropertyGroup` is not _that_ much of a pain, but I'd probably refactor it back into the previous one, anyways, on the next occasion 🤔

Lombiq.Analyzers/build/Lombiq.Analyzers.props
0liver: Mentioned the idea [on Teams here](https://teams.microsoft.com/l/message/19:525a1a3656c845ae92c5e2c0d3617cc7@thread.skype/1659557720569?tenantId=959ede5a-38e9-44e6-adc8-85553beb8220&groupId=d6a5d775-6fb9-41dc-a7bb-481d4f884046&parentMessageId=1659518116917&teamName=Lombiq%20Developers&channelName=General&createdTime=1659557720569).

New issue: https://lombiq.atlassian.net/browse/OSOE-167.

TestSolutions/Lombiq.Analyzers.PackageReference/IndirectPackageReference/IndirectPackageReference.csproj
0liver: > this project references Lombiq.Analyzers.PackageReference

I can't seem to find that reference - where is it?

Lombiq.Analyzers/Build.props
```diff
@@ -62,8 +66,22 @@
       <Analyzer Remove="@(Analyzer)" />
     </ItemGroup>
   </Target>
+  <!-- This target ensures, that LombiqNetAnalyzers is called during Build as well. This way the LombiqNetAnalyzers
```
0liver, line 68: Add an empty line between the `<Target>`s as everywhere else in the file.

TestSolutions/Directory.Build.props
```diff
@@ -0,0 +1,7 @@
+<Project>
```
0liver, line 3: Lets add a comment here along these lines:

> This file is needed, with properties or without, to prevent importing any Directory.Build.props higher up in the directory tree.

0liver (changes requested):

TestSolutions/Lombiq.Analyzers.PackageReference/IndirectPackageReference/IndirectPackageReference.csproj
sarahelsaig: WTF, it vanished!! 😱

sarahelsaig: Ok, now it's fixed. Updated related documentation too.

Invoke-BeforePack.ps1
sarahelsaig: We can't drop it, because of the other half of the comment (NuGet's validation). So yeah apparently it doesn't help with NU5100 warning, I guess I forgot to clean up the comment. I did so now.

I dread to touch this file, because if I mess something up I won't be able to see for myself. I'd have to wait and ask Zoltán if he got one of these accursed mails or not.
image
What I have above I reached purely by trial and error using this bs feedback. So I know it works and it's necessary. Don't know why, because NuGet itself is an opaque web service and web2 was a mistake.

Lombiq.Analyzers/Build.props
sarahelsaig: Well, as long as I don't have to file a bug with MS or OC I think I should be okay. :)

https://youtrack.jetbrains.com/issue/RIDER-80579/csproj-schema-doesnt-recognize-Condition-for-RunAnalyzersDuringBuild-property-specifically

Invoke-BeforePack.ps1
0liver: I feel your pain! I just saw that you're copying the .dll and .pdb files into _lib_ directly, **not** into any **framework-specific folder** under _lib_, which is what the link above suggests. From all the documentation on NU5100 says on the topic, that should fix the warning which would be nice™! I know you're faaar beyond nice with this issue, but suppressing warnings when we don't have to wouldn't be good in the long run.

I'm trying to fix this right now.

Lombiq.Analyzers/Build.props
0liver: Now, that we've identified this behavior as a tooling bug, let's move the property back where it belongs.

sarahelsaig: Okay. Added the issue link to the comment as well.

Invoke-BeforePack.ps1
0liver: Turns out that removing the .dll files from _\bin_ and _\obj_ folders fixes the NU5100 warning, so I was able to remove that from the GHA `publish-nuget` action. 🎉 

Now, creating a symbol package for a package that doesn't contain any .dll tripped the Symbol package validation process at NuGet.org:
image

This could be fixed by not publishing a symbol package in the first place. Which seems to make sense, given that `Lombiq.Analyzers` doesn't contain any code.

To support that in our publish workflow, we'd need a new parameter along the lines `SkipSymbolPackage` that would default to `false`, which we could set to `true` for the Analyzers package.

What do you think @DAud-IcI and @Piedone?

Piedone: I'm OK with that, just make sure to document in the YAML of this project why we're doing this.

0liver: That worked: https://github.com/Lombiq/.NET-Analyzers/runs/7672221377.

0liver (approved):

PR #58, opened 2022-05-19
CARE-228: Update Microsoft.CodeAnalysis.CSharp.CodeStyle and remove dependency on netstandard

0liver:
CARE-228


Piedone:
As usual, open a PR for this in OSOCE too.

0liver:
Of course! Why wouldn't I?

Luko6 (approved):

PR #63, opened 2022-05-29
OSOE-124: Fixing VSTHRD111 and .NET SDK code analysis rules

Piedone:
OSOE-124

### DIFF
--- a/general.ruleset
+++ b/general.ruleset
@@ -305,6 +305,7 @@
     <Rule Id="CA5403" Action="Warning" />
   </Rules>
   <Rules AnalyzerId="Microsoft.VisualStudio.Threading.Analyzers" RuleNamespace="Microsoft.VisualStudio.Threading.Analyzers">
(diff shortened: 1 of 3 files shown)


general.ruleset
```diff
@@ -429,4 +430,4 @@
     <Rule Id="SX1101" Action="Warning" />
     <Rule Id="SX1309S" Action="Warning" />
   </Rules>
-</RuleSet>
+</RuleSet>
```
0liver, line 433: Please don't remove empty lines at the end of files.

Piedone: The VS graphical editor for rulesets removes it. Just as we changed the file name casing due to it constantly doing that, I wouldn't keep re-adding this just for the sake of it.

0liver: I see. Makes sense, although it also makes me sad.

Piedone: Me too.

sarahelsaig (approved):

PR #80, opened 2022-12-23
CARE-310: Update Microsoft.Net.Compilers.Toolset to fix .NET Fx build

0liver:
[CARE-310](https://lombiq.atlassian.net/browse/CARE-310)


dministro (changes requested):
No PR for this in superproject (OSOCE)

dministro (approved):

PR #93, opened 2023-10-10
OSOE-179: Migrate to .editorconfig from ruleset files

### LINKED ISSUE #32: Migrate to .editorconfig from ruleset files (OSOE-179)
This is only suitable once .editorconfig has everything ruleset files provide already:
- IDE support, i.e. you don't have to manually add every rule but can just select what to do for each rule. This is now done with VS 2022. However, it only works for files under the solution, so you can't edit under the *tools* folder.
- [Ability to extend another .editorconfig file](https://github.com/editorconfig/editorconfig/issues/236). This is only interesting for overriding our default rules though (as we explain [here](https://github.com/Lombiq/.NET-Analyzers/blob/dev/Lombiq.Analyzers/Docs/ConfiguringAnalyzers.md#overriding-editorconfig-rules)). It would also help to get rid of the editorconfig copy to the solution root. However, perhaps the `<EditorConfigFiles>` could help us, see [this blog post](https://www.meziantou.net/sharing-coding-style-and-roslyn-analyzers-across-projects.htm). An alternative would be https://github.com/dotnet/roslyn/issues/19028#issuecomment-843571161.
- [There's also `.globalconfig`](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#global-analyzerconfig), but that's only for analyzer configs, not edit configuration like charset or guidelines. See https://github.com/dotnet/roslyn/issues/47707.

Tasks:
- [x] Migrate the config files, see the [official docs](https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2022#convert-an-existing-ruleset-file-to-editorconfig-file).
- [x] Perhaps we need to have use case-specific packages, e.g. `Lombiq.Analyzers.OrchardCore"?
- [x] Update docs. Search for ".editorconfig" and "ruleset".
- [x] Add docs on migration, including converting ruleset files.
- [x] Check if we need `<developmentDependency>true</developmentDependency>` in the nuspec template. The analyzer packages we use do this. Check the generated `<PackageReference>` after a new install and change the docs, including release notes, and consuming projects if necessary.
- [x] [Publish new NuGet package](https://github.com/Lombiq/.NET-Analyzers#publishing-a-new-nuget-package). This will need to be a major release. Add migration docs to the release.

Piedone:
Fixes #32
[OSOE-179](https://lombiq.atlassian.net/browse/OSOE-179)

**Release notes, to be copied to the GitHub release:**

### TL;DR upgrade instructions

If you used Lombiq .NET Analyzers in an Orchard Core app without overriding any of its configuration, then you have to change its package reference like this:

```diff
-<PackageReference Include="Lombiq.Analyzers" Version="x.y.z">
+<PackageReference Include="Lombiq.Analyzers.OrchardCore" Version="4.0.0">
```

Or if you used it as a Git submodule:

```diff
-<Import Project="tools/Lombiq.Analyzers/Lombiq.Analyzers/Build.props" />
+<Import Project="tools/Lombiq.Analyzers/Lombiq.Analyzers.OrchardCore/Build.props" /> 
```

If you have any other use case, then read on.

### Breaking changes

This release upgrades how analyzer configuration is stored. If you've overridden some analyzer configuration, then you'll need to migrate those.

Previously, editor configuration like charset and tab width was in an _.editorconfig_ file, together with some compatible analyzer rule configuration, and the rest of the analyzer rules were configured in RuleSet files. This release moves analyzer configuration to the currently recommended [_.globalconfig_](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#global-analyzerconfig) files, while keeping only editor configuration in the _.editorconfig_ file.

Furthermore, to better aid using `Lombiq.Analyzers` in any .NET project, regardless if it's an Orchard Core app or else, we've split the project into multiple, use case-specific projects/packages.

#### Select the `Lombiq.Analyzers` package suitable for you in any case

Previously, there was only a single `Lombiq.Analyzers` project/package that contained configuration for general .NET, Orchard Core, Orchard 1 (.NET Framework), and Visual Studio Extension projects. By default, this provided configuration for Orchard Core projects. If you wanted to use it in any other kind of project, you needed to jump through hoops.

Now, there are multiple projects/packages, all specific to a given use case. To get the most suitable configuration for your project, you just need to select the most applicable one. Please check out the [updated docs](https://github.com/Lombiq/.NET-Analyzers/blob/dev/Lombiq.Analyzers/Docs/AddingAnalyzers.md) to see how to add the one suitable for you.

#### If you haven't overridden any analyzer configuration and just used the project as is

Nothing else to do.

##### If you have overridden configuration coming from a RuleSet file:

RuleSets are now deprecated and you need to use _.globalconfig_ files for analyzer configuration instead.

1. Open the RuleSet file in a text editor, NOT in the Visual Studio GUI editor.
2. Remove the `<Include>` element that loads the default RuleSet file from `Lombiq.Analyzers` so only the overriding rules will be converted.
3. Follow [the official documentation](https://learn.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2022#convert-an-existing-ruleset-file-to-editorconfig-file) to convert the file into an _.editorconfig_ file. Save the file with the _.globalconfig_ name, NOT _.editorconfig_.
4. Put the file into the root of your solution if you want it to be applied to the whole solution. For more involved configuration options, see the [updated docs](https://github.com/Lombiq/.NET-Analyzers/blob/dev/Lombiq.Analyzers/Docs/ConfiguringAnalyzers.md#how-to-override-analyzer-configuration-globally).

##### If you have overridden configuration coming from the _.editorconfig_ file:

The _.editorconfig_ file now only stores editor-related configuration, like charset and tab settings. All analyzer configuration, also what was formerly in the _.editorconfig_ file, is now in _.globalconfig_ files. Follow the [updated docs](https://github.com/Lombiq/.NET-Analyzers/blob/dev/Lombiq.Analyzers/Docs/ConfiguringAnalyzers.md#how-to-override-analyzer-configuration-globally) to override analyzer configuration from _.globalconfig_ files. Overriding editor-related configuration with _.editorconfig_ files works the same as before.

[OSOE-179]: https://lombiq.atlassian.net/browse/OSOE-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Analyzers/Docs/AddingAnalyzers.md
```diff
@@ -1,8 +1,36 @@
...
+- [`Lombiq.Analyzers.NetFx`](../../Lombiq.Analyzers.NetFx/Readme.md): General .NET Framework projects.
+- [`Lombiq.Analyzers.OrchardCore`](../../Lombiq.Analyzers.OrchardCore/Readme.md): [Orchard Core](https://orchardcore.net/) projects.
+- [`Lombiq.Analyzers.Orchard1`](../../Lombiq.Analyzers.Orchard1/Readme.md): [Orchard 1](https://orchardcore.net/orchardcms) projects.
+- [`Lombiq.Analyzers.VisualStudioExtension`](../../Lombiq.Analyzers.VisualStudioExtension/Readme.md): Visual Studio extension projects.
+
+You only need to reference a single project; e.g., even though Orchard Core apps are also .NET apps, you only need to use `Lombiq.Analyzers.OrchardCore` for them.
+
+## How to add the analyzers to SDK-style projects from NuGet
+
+If you don't want to stay on the cutting-edge version, nor do you intend to contribute to Lombiq .NET Analyzers, you can use one of the NuGet packages. Install the package suitable for your project, as selected above. Check for the latest version number [on NuGet](https://www.nuget.org/packages/Lombiq.Analyzers/).
```
BenedekFarkas:
```suggestion
If you don't want to stay on the cutting-edge version, nor do you intend to contribute to Lombiq .NET Analyzers, you can use one of the NuGet packages. Install the package suitable for your project, as described above. Check for the latest version number [on NuGet](https://www.nuget.org/packages/Lombiq.Analyzers/).
```

Lombiq.Analyzers/Docs/AddingAnalyzers.md
```diff
@@ -47,39 +67,35 @@ For at least Visual Studio and JetBrains Rider you don't need any further setup
...
-    </PackageReference>
-```
+## How to add the analyzers to individual non-SDK-style .NET Framework projects (not solutions)
 
-The `<PrivateAssets>all</PrivateAssets>` is necessary to prevent the analyzers "leaking" into other projects that may consume yours.
+The key difference compared to .NET projects is that for non-SDK-style .NET Framework projects, you need to add Lombiq .NET Analyzers to each project.
 
-You can also add the package to all projects inside a folder at once [from a _Directory.Build.props_ file](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build).
+1. Same as above - add the .NET-Analyzers repository as a Git submodule to your repository or add the NuGet package in the next step.
+2. Create a _Directory.Build.props_ file in every project folder you want to target, next to the _.csproj_ file with the following content, either referencing `Lombiq.Analyzers.NetFx` or `Lombiq.Analyzers.Orchard1`:
```
BenedekFarkas: If all the projects that you want analyzed have a common ancestor folder that only contains the projects you want analyzed, then you can get away with a single _Directory.Build.props_ file here too, right?

Lombiq.Analyzers/Lombiq.Analyzers.VisualStudioExtension.globalconfig
BenedekFarkas, line 1: Isn't this a leftover duplicate of [Lombiq.Analyzers.VisualStudioExtension/Lombiq.Analyzers.VisualStudioExtension.globalconfig](https://github.com/Lombiq/.NET-Analyzers/pull/93/files#diff-ea78335fdbc3ac7ffd79417cdf890280fb782d659fb5dd7987a0fdaf5dac7cbe)?

BenedekFarkas (changes requested):

Lombiq.Analyzers.NetFx/NuGetIcon.png
BenedekFarkas, line 1: NuGetIcon.png and Readme.md must duplicated between the analyzer sub-projects, right? I was thinking that it would be nicer to have them in a common folder one level above and reference them from there, but then the path won't match up for the individual NuGet packages, I guess. Maybe a copy task in the sub-projects' csproj to copy them to the right place?

Lombiq.Analyzers.Orchard1/Lombiq.Analyzers.Orchard1.globalconfig
BenedekFarkas, line 1: Are the new .globalconfig files together completely supersede Lombiq.Analyzers/.editorconfig? I'm asking, because there are a lot more rules deleted from the latter compared to what's in the new files.

BenedekFarkas (changes requested):

Lombiq.Analyzers/Docs/AddingAnalyzers.md
Piedone: Yep. This is here because in O1 you usually don't have that, but I've clarified this.

Lombiq.Analyzers/Lombiq.Analyzers.VisualStudioExtension.globalconfig
Piedone: It is.

Lombiq.Analyzers.NetFx/NuGetIcon.png
Piedone: The Readmes are unique to each project.

NuGetIcon does need to be duplicated in all submodules of OSOCE (for the NuGet publishing to work independently in each of their repos) but it indeed doesn't necessarily need to be duplicated in all projects of these submodules. However, this is done automatically by [this target](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/blob/dev/Directory.Build.props#L17).

We could optimize this to only have a single copy per submodule, but that seems like a lot of work and attention needed for muli-project repos just to save a copy of instances of a 5 KB file (there wouldn't be any other benefit than storage; we already have DRY with the OSOCE target).

Lombiq.Analyzers.Orchard1/Lombiq.Analyzers.Orchard1.globalconfig
Piedone: They overlap but .editorconfig is still needed. Editor-specific config, like charset, guidelines, tabs, need to be in .editorconfig (since such configs aren't supported by .globalconfig, and non-.NET editors only understand .editorconfig). Analyzer configs can be in .editorconfig but the recommended way is to put them into .globalconfigs, since on contrary to .editorconfig those files can be referenced in a csproj instead of having to be copied to the folder structure.

So, editor configs should be in .editorconfig, analyzer configs in .globalconfigs.  I thus moved all the analyzer configs from the .editorconfig to the main Lombiq.Analyzers.globalconfig, cleaned it up, as well as added the analyzer configs migrated from .ruleset files there.

Lombiq.Analyzers.NetFx/NuGetIcon.png
BenedekFarkas: OK - BTW I meant to write License.md instead of Readme, but it's the same story as NuGetIcon.png (or even less common, because the year might be different).

Lombiq.Analyzers.Orchard1/Lombiq.Analyzers.Orchard1.globalconfig
BenedekFarkas: OK, thanks! What caught my eye at first is how much stuff was deleted from `.editorconfig`, but not so much was added to the subfolders' `.globalconfig` files. But I didn't look at the root `.globalconfig` (`Lombiq.Analyzers/Lombiq.Analyzers.globalconfig`), which is mostly the successor of `Lombiq.Analyzers/.editorconfig`, so I understand now!

BenedekFarkas (approved):

PR #94, opened 2023-10-18
HAST-327: New instructions about nested Directory.Build.props usage

sarahelsaig:
[HAST-327](https://lombiq.atlassian.net/browse/HAST-327)

[HAST-327]: https://lombiq.atlassian.net/browse/HAST-327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Analyzers/Docs/ConfiguringAnalyzers.md
+++ b/Lombiq.Analyzers/Docs/ConfiguringAnalyzers.md
@@ -21,6 +21,18 @@ This will completely disable code analysis packages. To also disable .NET SDK an
 dotnet_analyzer_diagnostic.category-Style.severity = none
 ```
 
+Note that MSBuild only loads in the closest _Directory.Build.props_ file to the project being built. So if you also used a _Directory.Build.props_ file to enable _Lombiq.Analyzers_ in a higher directory, it is now overridden. This can be a problem if you rely on `Lombiq.Analyzers` to set up compiler properties such as `<LangVersion>`. Put this into the _Directory.Build.props_ file instead:
+
+```xml
+<Project>
+    <PropertyGroup>
+        <RunCodeAnalysis>false</RunCodeAnalysis>
+    </PropertyGroup>
+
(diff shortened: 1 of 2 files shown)


Lombiq.Analyzers/Docs/ConfiguringAnalyzers.md
```diff
@@ -20,6 +20,18 @@ This will completely disable code analysis packages. To also disable .NET SDK an
 dotnet_analyzer_diagnostic.category-Style.severity = none
 ```
 
+Note that MSBuild only loads in the closest _Directory.Build.props_ file to the project being built. So if you also used a _Directory.Build.props_ file to enable _Lombiq.Analyzers_ in a higher directory, it is now overridden. This can be a problem if you rely on `Lombiq.Analyzers` to set up compiler properties such as `<LangVersion>`. Put this into the _Directory.Build.props_ file instead:
```
Piedone: So do we actually not need `<Analyzer Remove="@(Analyzer)" />` for the full analyzer removal, but an empty Directory.Build.props would suffice? If yes, please update the docs above too.

Piedone (changes requested):

Lombiq.Analyzers/Docs/ConfiguringAnalyzers.md
sarahelsaig: That wouldn't work if you include the project from NuGet package.

Piedone: What do you mean? If you do that then the package reference is either coming from a Directory file too, in which case it would work. Or is added directly to the csproj, but then you'd just remove the package, or use `<RunCodeAnalysis>false</RunCodeAnalysis>` too.

Piedone (approved):

PR #96, opened 2023-12-30
OSOE-760: Disable SA1518 since it's already covered by S113

0liver:
- warns on lacking newline at end of file

### DIFF
--- a/Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
+++ b/Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
@@ -649,6 +649,7 @@ dotnet_diagnostic.SA1503.severity = none
 dotnet_diagnostic.SA1512.severity = none
 dotnet_diagnostic.SA1515.severity = none
 dotnet_diagnostic.SA1516.severity = none
+dotnet_diagnostic.SA1518.severity = none
 dotnet_diagnostic.SA1519.severity = none
 dotnet_diagnostic.SA1601.severity = none
 dotnet_diagnostic.SA1602.severity = none


Piedone:
Thanks for the contrib! Don't worry about the missing parent PR failing check, I'll add one.

Any particular reason why you didn't choose to keep the StyleCop rule instead?

0liver:
Thanks! I had already forgotten that a parent PR is needed 😉

As for the choice of rule to keep - I think I liked the warning message better from `S113`, but looking at the great rules provided by StyleCop, keeping `SA1518` instead seems like a good idea, too. Would you like me to switch them around?

Piedone:
No, it's OK then. The StyleCop rule [also requires some configuration](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#lines-at-end-of-file) so that makes things easier. Interestingly enough, we didn't actually add any config but it was still working, perhaps by some coincidence.

Piedone:
Thank you!

0liver:
It's been a pleasure! 😊

0liver:
There's another duplicate, `CA1711` and `S2344`:

image

Which one would you remove?

Piedone:
I'd need to dig into that, but the SDK one is usually a safer bet.

0liver:
> the SDK one is usually a safer bet.

To keep, I suppose?

---

I just discovered why picking S113 over SA1518 might have been the wrong choice - there's no quick-fix action available for the former, but there is for the latter, which is greatly appreciated when you want to clean up your solution one warning at a time 😉

Do you fancy another PR to invert the exclusion?

Piedone:
Yes, and sure, thanks.

0liver:
Here's the follow-up PR: https://github.com/Lombiq/.NET-Analyzers/pull/97.

PR #98, opened 2024-01-18
OSOE-751: Upgrade to Orchard Core 1.8

Psichorex:
[OSOE-751](https://lombiq.atlassian.net/browse/OSOE-751)

[OSOE-751]: https://lombiq.atlassian.net/browse/OSOE-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Psichorex:
S3604 - Set to none as it interferes with IDE0290
S2094 - Set to none as it interferes with IDE0079 which has a much broader usecase

SA1010 - Set to none as just doesn't make sense at all

MA0110 - Set to none as it is information level on default according to docs

Lombiq.Analyzers/AnalyzerPackages.props
```diff
@@ -6,12 +6,12 @@
   <ItemGroup>
     <AnalyzerPackage Include="AsyncFixer" Version="1.6.0"/>
     <AnalyzerPackage Include="DotNetAnalyzers.DocumentationAnalyzers" Version="1.0.0-beta.59"/>
-    <AnalyzerPackage Include="Meziantou.Analyzer" Version="2.0.85"/>
-    <AnalyzerPackage Include="Microsoft.CodeAnalysis.CSharp.CodeStyle" Version="4.7.0"/>
-    <AnalyzerPackage Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.4"/>
-    <AnalyzerPackage Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.7.30"/>
+    <AnalyzerPackage Include="Meziantou.Analyzer" Version="2.0.136"/>
```
Piedone: `2.0.139` is the latest now.

Lombiq.Analyzers/AnalyzerPackages.props
```diff
@@ -6,12 +6,12 @@
...
+    <AnalyzerPackage Include="Microsoft.CodeAnalysis.CSharp.CodeStyle" Version="4.8.0"/>
+    <AnalyzerPackage Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="8.0.0"/>
+    <AnalyzerPackage Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.8.14"/>
     <AnalyzerPackage Include="SecurityCodeScan.VS2019" Version="5.6.7"/>
-    <AnalyzerPackage Include="SonarAnalyzer.CSharp" Version="9.10.0.77988"/>
+    <AnalyzerPackage Include="SonarAnalyzer.CSharp" Version="9.17.0.82934"/>
```
Piedone: `9.19.0.840257 is the latest now.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
```diff
@@ -571,6 +574,7 @@ dotnet_diagnostic.S134.severity = warning
 dotnet_diagnostic.S1696.severity = warning
 dotnet_diagnostic.S1698.severity = warning
 dotnet_diagnostic.S1858.severity = warning
+dotnet_diagnostic.S2094.severity = none
```
Piedone, line 577: This is [Classes should not be empty](https://rules.sonarsource.com/csharp/RSPEC-2094/), while IDE0079 is [Remove unnecessary suppression](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0079). We need both. Why would they clash?

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
```diff
@@ -602,6 +606,7 @@ dotnet_diagnostic.S3376.severity = none
 dotnet_diagnostic.S3441.severity = warning
 dotnet_diagnostic.S3458.severity = none
 dotnet_diagnostic.S3532.severity = warning
+dotnet_diagnostic.S3604.severity = none
```
Piedone: Why should we disable this? https://rules.sonarsource.com/csharp/RSPEC-3604/

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
```diff
@@ -634,6 +639,7 @@ dotnet_diagnostic.S881.severity = warning
 # StyleCop.Analyzers rules
 dotnet_diagnostic.SA0001.severity = none
 dotnet_diagnostic.SA1009.severity = none
+dotnet_diagnostic.SA1010.severity = none
```
Piedone: Why is this not suitable? https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1010.md

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
```diff
@@ -539,6 +541,7 @@ dotnet_diagnostic.MA0101.severity = none
 dotnet_diagnostic.MA0102.severity = none
 dotnet_diagnostic.MA0105.severity = warning
 dotnet_diagnostic.MA0106.severity = none
+dotnet_diagnostic.MA0110.severity = none
```
Piedone: This is good to have as warning: https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0110.md. According to [the official docs](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators)(https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators), "Where possible, use source generated regular expressions instead of compiling regular expressions using the [RegexOptions.Compiled](https://learn.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regexoptions#system-text-regularexpressions-regexoptions-compiled) option."

I've checked out the analyzer and it seems to work well.

Piedone (changes requested):

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Psichorex: This was producing a controversial double trigger for me at one of the places. So bascially if I supressed S2094 on an empty class IDE0079 was raised. 

Btw why do we need S2094? We have tons of empty classes and thus we need to supress all of them. I don't see the usecase of it.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Psichorex: That was due to PCs and IDE0290 but it can be reverted to warning.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Psichorex: Because it produces warnings with the new array initializators. It's so stupid. So now `var intArray = [.. 1, 2, 3]` woull produce this warning because there is a space between `= [`

Psichorex: Also if you remove the space it looks utter wierd and would produce a different warning that `=` should be followed by a space.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Psichorex: image
Accoding to MA0009 this code should be compilant
image

But it triggers MA0110. Regex source generation requires the Regex to partial. I would have to then convert the class which was non-partial to partial to be able to use the regex. Do we want that?

Psichorex: If yes then it can be made a warning.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Piedone: Fair point, let's leave it.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Piedone: I see. Please open a StyleCop issue about this if there's not one already and link it in the code.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Piedone: The example you show is specifically what benefits from source generators from most, since it's `RegexOptions.Compiled`. The line you highlight isn't, this is important.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Psichorex: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3794

Piedone: Following https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3794#issuecomment-1925600645, please update StyleCop.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
```diff
@@ -639,6 +639,7 @@ dotnet_diagnostic.S881.severity = warning
 # StyleCop.Analyzers rules
 dotnet_diagnostic.SA0001.severity = none
 dotnet_diagnostic.SA1009.severity = none
+# SA1010 interferes with IDE0055, see https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3794
 dotnet_diagnostic.SA1010.severity = none
```
Piedone: Following https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3794#issuecomment-1925600645, please update StyleCop.

Piedone (changes requested):

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Piedone: So here I agreed with making `S2094` not be a warning. It should remain being turned off, with is corresponding `#pragma warning disable S2094` suppressions removed again.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Psichorex: COmmented and updated the package. It indeed solved it.

Lombiq.Analyzers/Lombiq.Analyzers.globalconfig
Psichorex: Removed the supressions of the classes. Set S2094 to none.

Piedone (approved):

PR #103, opened 2024-06-11
OSOE-852: Change NuGet metadata to use license expression

AydinE:
[OSOE-852](https://lombiq.atlassian.net/browse/OSOE-852)

Removed license url 
Because I was getting: The licenseUrl and license elements cannot be used together. 
See run: https://github.com/Lombiq/.NET-Analyzers/actions/runs/9469828261/job/26089385371#step:4:538 

[OSOE-852]: https://lombiq.atlassian.net/browse/OSOE-852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Analyzers.NetFx/Lombiq.Analyzers.NetFx.nuspec.template
+++ b/Lombiq.Analyzers.NetFx/Lombiq.Analyzers.NetFx.nuspec.template
@@ -6,7 +6,6 @@
         <title>Lombiq .NET Analyzers for .NET Framework</title>
         <authors>Lombiq Technologies</authors>
         <license type="expression">BSD-3-Clause</license>
-        <licenseUrl>https://aka.ms/deprecateLicenseUrl</licenseUrl>
         <icon>NuGetIcon.png</icon>
         <projectUrl>https://github.com/Lombiq/.NET-Analyzers/blob/dev/Lombiq.Analyzers.NetFx</projectUrl>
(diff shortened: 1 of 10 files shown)


Lombiq.Analyzers.NetFx/Lombiq.Analyzers.NetFx.nuspec.template
```diff
@@ -5,8 +5,7 @@
         <version>$version$</version>
         <title>Lombiq .NET Analyzers for .NET Framework</title>
         <authors>Lombiq Technologies</authors>
-        <license type="file">License.md</license>
-        <licenseUrl>https://aka.ms/deprecateLicenseUrl</licenseUrl>
+        <license type="expression">BSD-3-Clause</license>
```
Piedone, line 8: Are still both needed, the license in the template and the csproj?

AydinE: Ah, no it seems this is no longer needed.
Test run: https://github.com/Lombiq/.NET-Analyzers/actions/runs/9470163660
Test release of NetFx package without the license expression in the csproj: https://www.nuget.org/packages/Lombiq.Analyzers.NetFx/5.0.1-alpha.3.osoe-852

Have updated it for all of the packages now

